PR #3, opened 2019-01-09
Add support for the `call` instruction

fitzgen:
r? @alexcrichton

This should give you an idea of what implementing another instruction is like, and a general layout of where things are, and how tests work, etc.


alexcrichton:
Looks fantastic to me 👍

fitzgen:
Thanks for taking a look!

PR #4, opened 2019-01-10
Update and tweak Travis configuration

alexcrichton:
* Use `sccache` to cache both C++ and Rust compilations
* Drop the `cargo` cache in favor of `sccache`
* Run `cargo test --all` instead of just `cargo test` to ensure all
  tests are run
* Fix tests with the nightly branch of `wabt` where the syntax for
  `get_local` has been changed to `local.get`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,20 @@
 language: rust
 sudo: false
 
-cache:
-  - cargo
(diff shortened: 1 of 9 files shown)


alexcrichton:
Will do 👍

PR #7, opened 2019-01-10
Clean up some boilerplate during tests

alexcrichton:
* Defer `wasm2wat` and such until the test itself
* All tests simply execute a `run` function
* Return a `Result` from the `run` function to centralize error handling
* Work with in-memory buffers where possible instead of files

### DIFF
--- a/walrus-tests/tests/valid.rs
+++ b/walrus-tests/tests/valid.rs
@@ -1,24 +1,11 @@
-extern crate failure;
-extern crate parity_wasm;
-extern crate walrus;
-
 use std::env;
 use std::fs;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use walrus::dot::Dot;
 
-fn do_assert_valid(path: &Path) {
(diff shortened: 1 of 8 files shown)


walrus-tests/tests/valid.rs
```diff
@@ -30,21 +17,13 @@ fn do_assert_valid(path: &Path) {
     assert_eq!(local_funcs.len(), 1);
 
     let f = &local_funcs.first().unwrap();
-    if env::var("WALRUS_TESTS_DOT").is_err() {
-        let mut dot_path = PathBuf::from(path);
-        dot_path.set_extension("dot");
-        let mut dot_file = fs::File::create(dot_path).expect("should create dot file OK");
-        f.dot(&mut dot_file).expect("should generate dot file OK");
+    if env::var("WALRUS_TESTS_DOT").is_ok() {
+        let dot_path = wat.with_extension("dot");
```
fitzgen: TIL `with_extension`!

PR #8, opened 2019-01-11
Add support for `memory.size` instruction

alexcrichton:


src/module/functions/local_function/mod.rs
```diff
@@ -661,6 +669,10 @@ impl LocalFunction {
                 self.emit(elements::Instruction::Return)?;
                 Err(())
             }
+
+            fn visit_memory_size(&mut self, e: &MemorySize) -> Self::Return {
+                self.emit(elements::Instruction::CurrentMemory(e.memory.index() as u8))
```
alexcrichton: This is one part in particular I was curious about, how do we know the `.index()` value with no other contextual information? If we gc a memory out (or something like that), don't we need a context to renumber memories?

(this happens also when printing the function)

src/module/functions/local_function/context.rs
```diff
@@ -48,6 +49,9 @@ pub struct FunctionContext<'a> {
     /// The locals for this function's module.
     pub locals: &'a mut ModuleLocals,
 
+    /// The memories for this function's module.
+    pub memories: &'a ModuleMemories,
```
alexcrichton: I was a little uneasy about this as well, do you think it's possible pass down a whole `&mut Module` to house all this information?

fitzgen: I think that makes sense. I am a little concerned that it might cause us to run into borrow issues where we want part of the module shared and part exclusive, but it it works then might as well roll with it.

src/module/functions/local_function/mod.rs
fitzgen: Oh this `.index()` is not the wasm index! That method is from the `id-arena` crate, and is the index of the thing that the id is pointing at in the arena's underlying vec! Not what we want here :)

Instead, this should be using a `&IdsToIndices` to map from the id to the index in the wasm we will emit: https://github.com/rustwasm/walrus/blob/5af5b9ff77a3a4e8e45d6c0749c9a3b02c56b370/src/module/emit.rs#L72

Something like

```rust
let idx = self.indices.get_memory_idx(e.memory);
assert!(idx < 256);
self.emit(elements::Instruction::CurrentMemory(idx as u8));
```

fitzgen: The `id.index()` thing is used elsewhere (like `Dot` and `Display` impls) just for debugging purposes -- we could probably clean that up to be the `fmt::Debug` instead or something. Didn't realize how misleading this would be when I wrote that...

PR #13, opened 2019-01-14
Rewrite how visitors work

alexcrichton:
This is a large-scale refactoring changing how visitors work. The goal
here is to make writing visitors a bit more concise and future-proof.
The major changes here are:

* The `Visitor` trait now has a default implementation for all methods.
  All methods automatically recurse through the AST, by default
  following `Visit`.

* The `Visitor` no longer has an associated `Return` type.

* The `Visit` trait is implemented for a few more types.

* The `Visitor` trait has "base methods" for all sorts of IDs, including
  `FunctionId` and `TableId`. These base methods do no recurse at all,
  but allow hooking references to these types.

* Many visitors have been simplified and/or cleaned up with this new
  visitor pattern.

* The `EmitVisitor` type has been extracted to its own module as it will
  likely grow quite a lot. This module no longer uses the `Visitor`
  trait but instead uses exhaustive matching to ensure we don't forget
  to write down how to emit something.

One of the caveats I ran into was that there's actually cycles in the
wasm AST, notably from branches to their destination blocks. To handle
these cycles and prevent stack overflows this skips them by default when
visiting. Only a DAG-like structure is visited by default.

Currently this doesn't update the dot or display implementations for
functions, but it's hoped that we can start automatically deriving those
soon!

### DIFF
--- a/src/dot.rs
+++ b/src/dot.rs
@@ -1,9 +1,7 @@
 //! Utilities for emitting GraphViz dot files.
 
-use std::io::{self, Write};
-
 /// Render something into a GraphViz dot file.
 pub trait Dot {
-    /// Render as dot into the given writer.
-    fn dot(&self, out: &mut Write) -> io::Result<()>;
+    /// Render as dot into the given string.
+    fn dot(&self, out: &mut String);
 }
--- a/walrus-derive/src/lib.rs
+++ b/walrus-derive/src/lib.rs
@@ -30,45 +30,90 @@ pub fn walrus_expr(_attr: TokenStream, input: TokenStream) -> TokenStream {
     TokenStream::from(expanded)
 }
 
-fn get_enum_variants(input: &DeriveInput) -> Vec<syn::Variant> {
-    match &input.data {
-        syn::Data::Enum(en) => en.variants.iter().cloned().collect(),
+struct WalrusVariant {
+    syn: syn::Variant,
+    fields: Vec<WalrusFieldOpts>,
+}
+
+#[derive(Default)]
+struct WalrusFieldOpts {
+    skip_visit: bool,
+}
+
+fn get_enum_variants(input: &DeriveInput) -> Vec<WalrusVariant> {
+    let en = match &input.data {
+        syn::Data::Enum(en) => en,
         syn::Data::Struct(_) => {
             panic!("can only put #[walrus_expr] on an enum; found it on a struct")
         }
         syn::Data::Union(_) => {
             panic!("can only put #[walrus_expr] on an enum; found it on a union")
         }
-    }
+    };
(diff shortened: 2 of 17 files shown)


walrus-derive/src/lib.rs
```diff
@@ -30,45 +30,90 @@ pub fn walrus_expr(_attr: TokenStream, input: TokenStream) -> TokenStream {
...
-    match &input.data {
-        syn::Data::Enum(en) => en.variants.iter().cloned().collect(),
+struct WalrusVariant {
+    syn: syn::Variant,
+    fields: Vec<WalrusFieldOpts>,
+}
+
+#[derive(Default)]
+struct WalrusFieldOpts {
+    skip_visit: bool,
```
fitzgen: Have you seen https://crates.io/crates/darling btw?

It is pretty awesome for deriving your derives (or at least deriving parsing of derive options)

yo dawg

walrus-tests/tests/round_trip/unreachable_2.wat
```diff
@@ -10,5 +8,6 @@
 ;; CHECK: (module
 ;; NEXT:    (type (;0;) (func (result i32)))
 ;; NEXT:    (func (;0;) (type 0) (result i32)
-;; NEXT:      unreachable)
+;; NEXT:      unreachable
+;; NEXT:      i32.const 42)
```
fitzgen: And the idea is we want to write a pass specifically for trimming unreachable isntructions instead of mixing it into `emit`, right?

src/dot.rs
```diff
@@ -1,9 +1,7 @@
 //! Utilities for emitting GraphViz dot files.
 
-use std::io::{self, Write};
-
 /// Render something into a GraphViz dot file.
 pub trait Dot {
-    /// Render as dot into the given writer.
-    fn dot(&self, out: &mut Write) -> io::Result<()>;
+    /// Render as dot into the given string.
+    fn dot(&self, out: &mut String);
```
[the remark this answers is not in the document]
alexcrichton: IIRC yeah, I wasn't able to propagate failure out of a visitor easily so I just changed this to be infallible

walrus-tests/tests/round_trip/unreachable_2.wat
alexcrichton: Indeed!

walrus-derive/src/lib.rs
alexcrichton: This looks pretty slick! I'm having a bit of a tough time following it though. I think for now this is an internal-enough derive that we can be pretty lossy/strict/bad, but if we get too much more advanced let's switch to something like darling

PR #14, opened 2019-01-15
Move the IR display visitor into `#[walrus_expr]`

alexcrichton:
This should reduce the overhead when adding instructions and also ensure
that all instructions are numbered and such. A sample of the new
debugging output is:

            (func
    (;  0;)   (block
    (;  2;)     (local.set 0
    (;  1;)       (const 9)
                )
    (;  3;)     (loop
    (;  6;)       (br.if (;e0;)
    (;  5;)         (i32.eqz
    (;  4;)           (local.get 0)
                    )
                  )
    (; 10;)       (local.set 0
    (;  9;)         (i32.add
    (;  7;)           (local.get 0)
    (;  8;)           (const 1)
                    )
                  )
                )
    (; 11;)     (const 10)
              )
            )

The tests have been updated for the new output as well. Note that the
expression ids aren't included in the test output but are instead
stripped out, as it's not too critical to be matching them.

### DIFF
--- a/walrus-derive/src/lib.rs
+++ b/walrus-derive/src/lib.rs
@@ -7,24 +7,34 @@ use heck::SnakeCase;
 use proc_macro2::Span;
 use quote::quote;
 use std::iter::FromIterator;
-use syn::{parse_macro_input, DeriveInput};
+use syn::DeriveInput;
+use syn::ext::IdentExt;
+use syn::Error;
+use syn::{parse_macro_input, Ident, Result, Token};
+use syn::parse::{Parse, ParseStream};
+use syn::punctuated::Punctuated;
 
 #[proc_macro_attribute]
 pub fn walrus_expr(_attr: TokenStream, input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
 
-    let variants = get_enum_variants(&input);
+    let variants = match get_enum_variants(&input) {
+        Ok(v) => v,
+        Err(e) => return e.to_compile_error().into(),
+    };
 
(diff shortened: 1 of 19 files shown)


walrus-derive/src/lib.rs
```diff
@@ -54,42 +71,101 @@ fn get_enum_variants(input: &DeriveInput) -> Vec<WalrusVariant> {
...
+        impl Parse for Attr {
+            fn parse(input: ParseStream) -> Result<Self> {
+				let attr: Ident = input.parse()?;
+				if attr == "display_name" {
+                    input.parse::<Token![=]>()?;
+                    let name = input.call(Ident::parse_any)?;
+					return Ok(Attr::DisplayName(name))
+				}
+				if attr == "display_extra" {
+                    input.parse::<Token![=]>()?;
```
fitzgen: Looks like you mixed tabs and spaces here, so the indent is funky.

walrus-derive/src/lib.rs
```diff
@@ -559,3 +642,71 @@ fn create_matchers(variants: &[WalrusVariant]) -> impl quote::ToTokens {
...
+        });
+    }
+    quote! {
+        impl<'expr> Visitor<'expr> for crate::module::functions::DisplayExpr<'expr, '_> {
+            fn local_function(&self) -> &'expr crate::module::functions::LocalFunction {
+                self.func
+            }
+
+            fn visit_expr_id(&mut self, expr: &ExprId) {
+                self.expr_id(*expr)
```
fitzgen: Nice.

PR #16, opened 2019-01-16
Remove `ty` from `LocalGet` and `LocalSet`

### LINKED ISSUE #10: Remove `ty` from `LocalGet`
also `LocalSet`

alexcrichton:
This is auxiliary information available in other tables, so let's just
keep a `LocalId` there. This involved tweaking the emission code to
ensure the table of locals made its way all the way down to function emission.

Closes #10


alexcrichton:
Agreed about the side table!

PR #17, opened 2019-01-16
Add support for element segments and `CallIndirect`

alexcrichton:
Lots of stuff added here, sort of by accident:

* `CallIndirect` support
* Parsing and serializing the elements wasm section
* Liveness tracking of passive segments
* Liveness tracking through function tables of retained function
  pointers
* Fixed parsing of `Call` with zero return values

### DIFF
--- a/src/module/elements.rs
+++ b/src/module/elements.rs
@@ -3,7 +3,7 @@
 use crate::error::Result;
 use crate::module::emit::{Emit, IdsToIndices};
 use crate::module::functions::FunctionId;
-use crate::module::tables::TableId;
+use crate::module::tables::{ModuleTables, TableKind};
 use crate::module::Module;
 use crate::passes::Used;
 use id_arena::{Arena, Id};
@@ -23,16 +23,17 @@ pub struct Element {
 /// used as function pointers.
 #[derive(Debug, Default)]
 pub struct ModuleElements {
-    active: HashMap<TableId, Vec<Option<FunctionId>>>,
-    passive: Arena<Element>,
+    elements: Arena<Element>,
     index_to_element_id: HashMap<u32, ElementId>,
 }
 
 impl ModuleElements {
     /// Parses a raw was section into a fully-formed `ModuleElements` instance.
-    pub fn parse(module: &Module, section: &elements::ElementSection) -> Result<ModuleElements> {
-        let mut active = HashMap::new();
-        let mut passive = Arena::new();
+    pub fn parse(
+        module: &mut Module,
+        section: &elements::ElementSection,
+    ) -> Result<ModuleElements> {
+        let mut elements = Arena::new();
         let mut index_to_element_id = HashMap::new();
 
         for (i, segment) in section.entries().iter().enumerate() {
@@ -44,8 +45,8 @@ impl ModuleElements {
                         None => failure::bail!("invalid segment initialization"),
                     }
                 }
-                let id = passive.next_id();
-                let ret = passive.alloc(Element { members: list });
+                let id = elements.next_id();
+                let ret = elements.alloc(Element { members: list });
                 debug_assert_eq!(id, ret);
                 index_to_element_id.insert(i as u32, id);
                 continue;
@@ -55,7 +56,10 @@ impl ModuleElements {
                 .tables
                 .table_for_index(segment.index())
                 .ok_or_else(|| failure::format_err!("invalid table index"))?;
-            let list = active.entry(table).or_insert(Vec::new());
+            let list = match &mut module.tables.get_mut(table).kind {
+                TableKind::Function(list) => list,
+            };
+
             let offset = segment.offset().as_ref().unwrap();
             if offset.code().len() != 2 {
                 failure::bail!("invalid initialization expression");
@@ -82,8 +86,7 @@ impl ModuleElements {
         }
 
         Ok(ModuleElements {
-            active,
-            passive,
+            elements,
             index_to_element_id,
         })
     }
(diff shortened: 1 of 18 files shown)


src/module/elements.rs
```diff
@@ -0,0 +1,185 @@
...
+use parity_wasm::elements;
+use std::collections::HashMap;
+
+/// A passive element segment identifier
+pub type ElementId = Id<Element>;
+
+/// A passive segment which contains a list of functions
+#[derive(Debug)]
+pub struct Element {
+    members: Vec<FunctionId>,
```
fitzgen: Not a set because order is important, right?

Maybe we *also* want to maintain a set for if someone wants to add a new function, and can instead just reuse and existing member? Eh, probably not worth it till we need it...

src/module/elements.rs
```diff
@@ -0,0 +1,185 @@
...
+        let mut passive = Arena::new();
+        let mut index_to_element_id = HashMap::new();
+
+        for (i, segment) in section.entries().iter().enumerate() {
+            if segment.passive() {
+                let mut list = Vec::with_capacity(segment.members().len());
+                for &func in segment.members() {
+                    match module.funcs.function_for_index(func) {
+                        Some(id) => list.push(id),
+                        None => failure::bail!("invalid segment initialization"),
```
fitzgen: Maybe worth adding some `{}` containing the function index here for better debugging.

src/module/elements.rs
```diff
@@ -0,0 +1,185 @@
...
+            if offset.code().len() != 2 {
+                failure::bail!("invalid initialization expression");
+            }
+            match offset.code()[1] {
+                elements::Instruction::End => {}
+                _ => failure::bail!("invalid initialization expression"),
+            }
+            let offset = match offset.code()[0] {
+                elements::Instruction::I32Const(n) => n as usize,
+                _ => failure::bail!("invalid initialization expression"),
```
fitzgen: Maybe worth re-using our dummy local functions as init exprs stuff here (and cleaning it up some more). I expect that const expressions will be expanded and eventually include arbitrary arithmetic.

They can also use imported globals already, so this is technically incomplete (even though I haven't seen it in the wild yet).

src/module/functions/local_function/mod.rs
```diff
@@ -373,7 +377,30 @@ fn validate_instruction<'a>(
...
+                .module
+                .types
+                .type_for_index(*type_idx)
+                .ok_or_else(|| format_err!("invalid type index in `call_indirect`"))?;
+            let ty = ctx.module.types.types()[type_id].clone();
+            let table = ctx
+                .module
+                .tables
+                .table_for_index((*table_idx) as u32)
+                .ok_or_else(|| format_err!("invalid index in `call_indirect`"))?;
```
fitzgen: Ditto on including the index in the error message.

src/module/functions/local_function/context.rs
```diff
@@ -110,11 +110,16 @@ impl<'a> FunctionContext<'a> {
...
 
-    pub fn push_operands<E>(&mut self, types: &[ValType], exprs: &[E])
+    pub fn push_operands<E>(&mut self, types: &[ValType], exprs: &[E], expr: ExprId)
     where
         E: Copy + Into<ExprId>,
     {
-        impl_push_operands(&mut self.operands, types, exprs)
+        assert_eq!(types.len(), exprs.len());
+        if types.is_empty() && self.controls.len() > 0 {
+            self.add_to_current_frame_block(expr);
```
fitzgen: This is really nice :+1:

src/module/functions/mod.rs
```diff
@@ -336,6 +333,11 @@ impl Emit for ModuleFunctions {
             codes.push(elements::FuncBody::new(locals, instructions));
         }
 
+        assert_eq!(funcs.len(), codes.len());
+        if codes.len() == 0 {
```
fitzgen: `is_empty()`

I'm the walking talking clippy over here

src/passes/used.rs
```diff
@@ -36,95 +40,114 @@ impl Used {
...
+                        }
+                        FunctionKind::Import(i) => {
+                            stack.used.imports.insert(i.import);
+                        }
+                        FunctionKind::Uninitialized(_) => unreachable!(),
+                    }
                 }
-                FunctionKind::Import(i) => {
-                    used.imports.insert(i.import);
+                ItemToVisit::Table(t) => {
```
fitzgen: FWIW: we could also have two different stacks for tables and functions if you think that would be more concise than defining a "throw away" enum.

src/module/elements.rs
alexcrichton: Correct yeah, this is like a data segment which is a blob of bytes and can be copied to/from.

I'll rename this to have `Passive` in the name as I think that makes more sense

src/module/elements.rs
alexcrichton: Yeah I figured that this'll need to be shored up one way or another, but I wasn't sure where it would best fit today. It's slightly different from the globals side as that was just validation as opposed to extracting the actual value. Do you think we should have like a `const_eval` function which evaluates to a `ConstValue` where one variant is a global perhaps?

src/passes/used.rs
alexcrichton: Yeah it ended up being a bit smaller, so switched to that

src/module/elements.rs
alexcrichton: Hm actually not a huge fan of `Passive`, so left as-is.

src/module/elements.rs
alexcrichton: I'll open up a follow-up issue.

PR #18, opened 2019-01-17
Add dedicated constant value processing

alexcrichton:
This commit unifies the paths in handling offsets in table elements as
well as initialization expressions in globals. A new `Const` type is
added and a `Const::eval` function is added for transforming an
`InitExpr`.

Notably `Const::Global` exists for non-constant values, and although we
don't parse it just yet support should be plumbed through.

### DIFF
--- a/src/const_value.rs
+++ b/src/const_value.rs
@@ -1,0 +1,52 @@
+//! Handling wasm constant values
+
+use crate::error::Result;
+use crate::ir::Value;
+use crate::module::emit::IdsToIndices;
+use crate::module::globals::GlobalId;
+use failure::bail;
+use parity_wasm::elements::{self, Instruction};
+
+/// A constant which is produced in WebAssembly, typically used in global
+/// initializers or element/data offsets.
+#[derive(Debug)]
+pub enum Const {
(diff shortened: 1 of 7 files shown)


src/const_value.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::ir::Value;
+use crate::module::emit::IdsToIndices;
+use crate::module::globals::GlobalId;
+use failure::bail;
+use parity_wasm::elements::{self, Instruction};
+
+/// A constant which is produced in WebAssembly, typically used in global
+/// initializers or element/data offsets.
+#[derive(Debug)]
+pub enum Const {
```
alexcrichton: I'm realizing now that `Const` is a terrible name because the whole point of its existence here is that it may not be constant.

Perhaps `Initializer`?

fitzgen: `ConstExprValue`?

alexcrichton: Hm ok, I think I'll leave as-is for now and we can change it if we get too annoyed at it. It's not so far off because the global's value will be constant on import.... right? ish?

PR #23, opened 2019-01-22
Add support for memory instructions

alexcrichton:
Curious on thoughts for the `LoadKind` and `StoreKind` enums, I'm not sure if those are the most useful or best designed here.

### DIFF
--- a/src/ir/mod.rs
+++ b/src/ir/mod.rs
@@ -301,6 +301,36 @@ pub enum Expr {
         /// The number of pages to grow by.
         pages: ExprId,
     },
+
+    /// Loading a value from memory
+    Load {
+        /// The memory we're loading from.
+        memory: MemoryId,
+        /// The kind of memory load this is performing
+        #[walrus(skip_visit)]
+        kind: LoadKind,
+        /// The alignment and offset of this memory load
+        #[walrus(skip_visit)]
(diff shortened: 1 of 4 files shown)


src/ir/mod.rs
```diff
@@ -478,6 +508,57 @@ pub enum UnaryOp {
...
 }
 
+/// The different kinds of load instructions that are part of a `Load` IR node
+#[derive(Debug, Copy, Clone)]
+#[allow(missing_docs)]
+pub enum LoadKind {
+    // TODO: much of this is probably redundant with type information already
+    // ambiently available, we probably want to trim this down to just "value"
+    // and then maybe some sign extensions. We'd then use the type of the node
+    // to figure out what kind of store it actually is.
```
fitzgen: I am in favor of de-duping this a bit and giving the relevant load memory instructions a `SignExtKind = Unsigned | Signed`. Maybe also using `ValType` with `Option<(u8, SignExtKind)>` where `u8` is the width of the narrow load. But def +1 to deduplicating `Blah_U` and `Blah_S`.

alexcrichton: Ok I'm gonna go with a `I32_8 { sign_extend: bool }` for now, but I'm gonna open an issue to talk about this a bit more too

PR #24, opened 2019-01-23
Add a suite of spec tests

alexcrichton:
The WebAssembly organization on GitHub has a `spec-tests` repository
which contains a collection of tests for WebAssembly implementations.
These tests are quite comprehensive and a great way to ensure good
coverage within our own engine of WebAssembly features, even as they're
added to the spec over time.

This commit adds a submodule to the spec-tests repository and then hooks
up a suite of tests to run them. To "run" a spec test means:

* First, the `wabt` toolkit's `wast2json` tool is executed, performing
  two operations. First, as the name implies, it parses the `*.wast`
  file and emits a `*.json` file describing the test. This `*.json` file
  is much easier to parse for us thatn `*.wast`. Second the tool will
  extract a number of `*.wasm` files from the test, emitting them to the
  filesystem.

* Next we look at all the test directives in the JSON file. For anything
  that's supposed to be an invalid/malformed module we ensure that our
  parsing fails (as our parsing currently bundles validation).

* For all directives which are a valid wasm test, we parse the
  referenced file and then serialize it back out, round-tripping the
  wasm file through `walrus`.

* Finally we execute `wabt`'s tool `spectest-interp`. This takes in the
  JSON file as input and then executes all of the tests, presumably in
  its built-in interpreter. This hopefully ensures that we actually
  emitted the correct wasm, as the tool is ingesting wasm that walrus
  produced.

Turns out we were already passing all the spec tests basically (yay!)
except for a few more validations added here:

* Exports are now an `Arena` instead of an `ArenaSet` to generate an
  error when two exports have the same name (and the same type).
  Previously they silently overrode each other.

* Alignment on loads/stores are validated to be less than the width of
  the load/store operation.

* The signature of the start function is checked.

* Added support for the sign-extension-ops proposal

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "walrus-tests/tests/spec-tests"]
+	path = walrus-tests/tests/spec-tests
+	url = https://github.com/WebAssembly/testsuite
--- a/src/ir/mod.rs
+++ b/src/ir/mod.rs
@@ -506,6 +506,12 @@ pub enum UnaryOp {
     I64ReinterpretF64,
     F32ReinterpretI32,
     F64ReinterpretI64,
+
+    I32Extend8S,
+    I32Extend16S,
+    I64Extend8S,
+    I64Extend16S,
+    I64Extend32S,
 }
 
 /// The different kinds of load instructions that are part of a `Load` IR node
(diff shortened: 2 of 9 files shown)


alexcrichton:
Note that this builds on https://github.com/rustwasm/walrus/pull/23, so the first commit can be ignored

PR #26, opened 2019-01-24
Switch to using `wasmparser` to parse wasm files

alexcrichton:
This commit moves our parser of wasm files from `parity-wasm` to the
`wasmparser` crate. The `wasmparser` crate provides a few key
advantages:

* Byte positions. These will be critical for parsing and updating DWARF
  debug information, so we need to ensure we have access to this
  information which `parity-wasm` doesn't currently expose.

* Speed. The `parity-wasm` crate isn't necessarily built for our use
  case as it sort of parses into its own IR-like representation which we
  then translate to our own IR. The `wasmparser` crate is much more
  "raw" and basically eschews all allocation to us, allowing us to parse
  much more quickly.

The `parity-wasm` crate currently supports a few more wasm proposals
than `wasmparser` like SIMD and bulk memory operations, so support for
these has temporarily been lost. In `walrus` we didn't fully support
them anyway though :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ id-arena = "2.0.1"
 parity-wasm = "0.35.6"
 petgraph = "0.4.13"
 log = "0.4"
-wasmparser = "0.23"
+wasmparser = "0.24"
 
 [dependencies.walrus-derive]
 path = "./walrus-derive"
@@ -22,6 +22,3 @@ members = [
     "./walrus-tests",
     "./walrus-tests-utils",
 ]
-
-[patch.crates-io]
-wasmparser = { git = 'https://github.com/alexcrichton/wasmparser.rs', branch = 'changes' }


alexcrichton:
Note that I did change a few things in `wasmparser`, so this currently depends on https://github.com/yurydelendik/wasmparser.rs/pull/83

fitzgen:
+cc @yurydelendik

src/module/producers.rs
```diff
@@ -113,3 +90,29 @@ impl ModuleProducers {
         })
     }
 }
+
+impl Module {
+    /// Parse a producers section from the custom section payload specified.
+    pub(crate) fn parse_producers_section(
+        &mut self,
+        mut data: wasmparser::BinaryReader,
```
fitzgen: Seems like we should upstream a `ProducersSectionReader` or whatever to `wasmparser`

PR #28, opened 2019-01-24
demo: Parallel parsing and validation 

alexcrichton:
This commit is a demonstration of using `rayon` to parallelize parsing
of the input wasm file as well as a simple pass, like validation. The
salient points here are:

* Function parsing is split in two, the first part being assigning a
  `LocalId` for all locals, and the second being actually parsing all of
  the function bodies. This allows us to do a bunch of mutable work up
  front, followed by all the parallel work.

* The `id-arena` crate has been patched to have a `rayon` feature so
  it has a `par_iter` method to iterate over functions in parallel.

Other than that, the transition is about as easy as you'd expect using
rayon!


alexcrichton:
This builds on https://github.com/rustwasm/walrus/pull/27 so only the last commit is really relevant here, but I'm curious as to your thoughts on this! I think that we may want to make "pass over the module in parallel" a bit easier/ergonomic to do, but I figure we can take care of that later

Also the `id-arena` changes look like this https://github.com/alexcrichton/id-arena/commit/44906ca325156e64b510ccbdc767572b404c562a

src/module/functions/mod.rs
```diff
@@ -212,6 +219,11 @@ impl Module {
         }
         let num_imports = self.funcs.arena.len() - (amt as usize);
 
+        // First up serially create corresponding `LocalId` instances for all
+        // functions as well as extract the operators parser for each function.
+        // This is pretty tough to parallelize, but we can look into it later if
+        // necessary and it's a bottleneck!
```
fitzgen: We would need to either make each function have a separate arena of locals, or put the local arena in a mutex (which would kind of defeat the purpose)

alexcrichton:
> building up each function body's bytes in parallel and then concatenating them together serially at the end.

Agreed! That's actually my next commit :)

(ish)

src/module/functions/mod.rs
alexcrichton: Yeah I think if this becomes a bottleneck we have a few ways to handle it. We need to persist the mapping of index to id beyond function parsing to handle the local names section, but I think we could have "bulk insert" operations where the arena is built locally per-function and then inserted into a map at the end.

I think we can actually go a pretty long way avoiding any synchronization beyond rayon, and ideally adding rayon doesn't add too much complexity in this crate

PR #35, opened 2019-01-28
Define a walrus-specific hasher for hash maps

alexcrichton:
One of the hottest parts of profiling `examples/round-trip.rs` on
`clang.wasm` is hashing, presumably with lots of lookups and stores. The
standard library hashing is known to not be optimized for speed by
default, but that's for DoS protection that we're not too concerned
about here.

Instead this commit defines a custom hasher specifically designed to
quickly hash `Id<T>` from the `id-arena` crate. Overally this improves
the runtime of `round-trip` on `clang.wasm` by 15%, reducing the runtime
from 1.9s to 1.6s.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1,0 +1,59 @@
+//! Walrus-specific hash maps and hash sets which typically hash much more
+//! quickly than libstd's hash map which isn't optimized for speed.
+
+use id_arena::Id;
+use std::collections::{HashMap, HashSet};
+use std::hash::{BuildHasher, Hasher};
+
+/// Type parameter for the hasher of a `HashMap` or `HashSet`
+#[derive(Default, Copy, Clone, Debug)]
+pub struct BuildIdHasher;
+
+/// Hasher constructed by `BuildIdHasher`, only ever used to hash an `Id`
+#[derive(Default, Copy, Clone, Debug)]
+pub struct IdHasher {
+    hash: u64,
+}
+
+pub type IdHashMap<K, V> = HashMap<Id<K>, V, BuildIdHasher>;
+pub type IdHashSet<T> = HashSet<Id<T>, BuildIdHasher>;
+
+impl BuildHasher for BuildIdHasher {
+    type Hasher = IdHasher;
+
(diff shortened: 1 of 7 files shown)


src/map.rs
```diff
@@ -0,0 +1,59 @@
...
+// ID. We can assume that none of the arena indexes are larger than 32-bit
+// because wasm can't encode more than 2^32 items anyway, so we can basically
+// just shift each item into the lower 32-bits of the hash. Ideally this'll end
+// up producing a very optimized version of hashing!
+//
+// To explore this a bit, see https://godbolt.org/z/QxkXgF
+impl Hasher for IdHasher {
+    fn write_u32(&mut self, amt: u32) {
+        self.hash <<= 32;
+        self.hash |= amt as u64;
```
fitzgen: Rather than always shifting the hash and oring the amt into it, can we just dedicate the upper half to the u32 and then just or the usize without any shifting? This should cut down on one shift (unless LLVM sees through it).

```rust
fn write_u32(&must self, amt: u32) {
    let amt = (amt as u64) << 32;
    self.hash |= amt;
}

fn write_usize(&mut self, amt: usize) {
    self.hash |= amt as u64;
}
```

Even if LLVM sees through everything, this is a little more clear to me as a reader.

src/map.rs
```diff
@@ -0,0 +1,59 @@
+// Walrus-specific hash maps and hash sets which typically hash much more
+// quickly than libstd's hash map which isn't optimized for speed.
```
fitzgen:
```
//!
```

src/map.rs
alexcrichton: 👍 sgtm

alexcrichton: Oh actually I remember now, I was hoping to slightly future-proof this because I think we may need to look into making the `usize` in the `Id` a `u32` for walrus specifically, our IR nodes are currently *massive* to the point that processing clang.wasm peaks at 2GB of memory for a 15MB wasm file, a nearly 150x blowup lol

PR #36, opened 2019-01-29
Implement support for all atomic instructions

alexcrichton:
This commit implements support for all atomic instructions in `walrus`,
adding definitions for all instructions proposed in the threads proposal
as well as hooking up validation to ensure their usage requires shared
memory.

There's not currently a working spec test for these unfortunately, but
this implementation passes against a modified version of the spec tests
currently upstream which wabt presently rejects due to instruction
renaming. We'll be sure to update to run the spec test once it's been
udpated!

### DIFF
--- a/src/ir/mod.rs
+++ b/src/ir/mod.rs
@@ -332,6 +332,74 @@ pub enum Expr {
         /// The value that we're storing
         value: ExprId,
     },
+
+    /// An atomic read/modify/write operation
+    AtomicRmw {
+        /// The memory we're modifying
+        memory: MemoryId,
+        /// The atomic operation being performed
+        #[walrus(skip_visit)]
+        op: AtomicOp,
+        /// The atomic operation being performed
+        #[walrus(skip_visit)]
+        width: AtomicWidth,
+        /// The alignment and offset from the base address
+        #[walrus(skip_visit)]
+        arg: MemArg,
+        /// The address at which to perform the operation
+        address: ExprId,
+        /// The value that's being used to modify what's at the address
+        value: ExprId,
+    },
+
+    /// An atomic compare exchange operation
+    Cmpxchg {
+        /// The memory we're modifying
+        memory: MemoryId,
+        /// The atomic operation being performed
+        #[walrus(skip_visit)]
+        width: AtomicWidth,
+        /// The alignment and offset from the base address
+        #[walrus(skip_visit)]
+        arg: MemArg,
+        /// The address at which to perform the operation
+        address: ExprId,
+        /// The value that's being used to modify what's at the address
+        expected: ExprId,
+        /// The value that's being used to modify what's at the address
+        replacement: ExprId,
+    },
+
+    /// The `atomic.notify` instruction to wake up threads
+    AtomicNotify {
+        /// The memory we're notifying through
+        memory: MemoryId,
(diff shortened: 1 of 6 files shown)


src/passes/validate.rs
```diff
@@ -63,6 +63,7 @@ pub fn run(module: &Module) -> Result<()> {
                 errs: &mut errs,
                 function,
                 local,
+                module,
```
alexcrichton: Just wanted to call out that I only realized after the fact that this was being shared concurrently between threads!

src/passes/validate.rs
```diff
@@ -199,12 +215,42 @@ impl<'a> Visitor<'a> for Validate<'a> {
...
+    }
+
+    fn visit_cmpxchg(&mut self, e: &Cmpxchg) {
+        self.require_atomic(e.memory, &e.arg, e.width.bytes());
+        e.visit(self);
+    }
+
+    fn visit_atomic_notify(&mut self, e: &AtomicNotify) {
+        self.require_shared(e.memory);
+        // TODO: should alignment or things be validated?
```
alexcrichton: FWIW I think the spec is somewhat unclear on this, so I figured we can just always add points later as necessary

src/ir/mod.rs
```diff
@@ -552,16 +620,25 @@ pub enum LoadKind {
...
+    I64_32 { kind: ExtendedLoad },
+}
+
+/// The kinds of extended loads which can happen
+#[derive(Debug, Copy, Clone)]
+#[allow(missing_docs)]
+pub enum ExtendedLoad {
+    SignExtend,
+    ZeroExtend,
+    ZeroExtendAtomic,
```
fitzgen: Huh, so there aren't any sign extending atomics?

src/module/functions/local_function/mod.rs
```diff
@@ -794,102 +831,298 @@ fn validate_instruction(
...
-            LoadKind::I32_16 { sign_extend: true },
-        )?,
-        Operator::I32Load16U { memarg } => load(
-            ctx,
-            memarg,
-            ValType::I32,
-            LoadKind::I32_16 { sign_extend: false },
-        )?,
-        Operator::I64Load8S { memarg } => load(
+        O::Nop => {}
```
fitzgen: Not convinced `Operator` -> `O` is really a win, but whatever...

src/ir/mod.rs
alexcrichton: I was also surprised when writing, but I couldn't find them [in the spec](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md). What I did find, however, was [this section](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#new-sign-extending-operators) "New Sign-extending Operators". I think that the CG must have decided that unsigned ops are the way to go from now on and sign extension should go through https://github.com/WebAssembly/sign-extension-ops/, dedicated instructions to sign extend values

src/module/functions/local_function/mod.rs
alexcrichton: I largely wanted to makeit so `cargo fmt` preserved reasonably readable code here and didn't spew all the parsing over hundreds of lines, but I don't think I was overly successful. I'll switch it back

PR #37, opened 2019-01-29
Implement some of bulk memory operations

alexcrichton:
This commit implements the memory-related pieces of the bulk memory
operations proposal. This doesn't currently implement the pieces for
tables because passive table elements depend on `ref.null` and
`ref.func`, both of which are still somewhat up in the air in the
reference types proposal. For now though we can get the benefit of
passive memory initialization as well as memory copies/fills/etc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ id-arena = { version = "2.1.0", features = ['rayon'] }
 leb128 = "0.2.3"
 log = "0.4"
 rayon = "1.0.3"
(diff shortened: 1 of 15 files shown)


fitzgen:
Would you like eyes on this now, or wait until upstream branches merge and CI is passing?

alexcrichton:
Let's wait for other CI, no rush here

alexcrichton:
Ok, should be good to go!

PR #38, opened 2019-01-31
VisitMut, FunctionBuilder, other conveniences

alexcrichton:
In transitioning some code in `wasm-bindgen` to using `walrus` I ended
up adding a number of types and things to `walrus`, notably:

* A `FunctionBuilder` type is now available to build a `LocalFunction`
  programmatically. This type has a associated function for each IR node
  (the arguments being the node's components) which is auto-generated.
  It then has special methods for handling blocks to ensure that block
  IDs can be generated ahead-of-time.

* A `VisitMut` and `VisitorMut` pair of traits is intended to mirror the
  `Visit` and `Visitor` traits, only be mutable versions. Currently they
  have a 1:1 correspondance except for `&` vs `&mut`. I'm not entirely
  sure how this will play out long-term, but I figure it's a good start.

  The most notable part here is `VisitMut for ExprId` which *removes* an
  expression from the arena (replacing it with a dummy expression),
  visits mutably recursively, and then re-inserts it back into the
  arena. This should work for now but may pose problems once we have a
  DAG of an IR instead of a tree.

* A few other conveniences for iterating and such have been added. As a
  public API I've gone ahead an made most of the fields of `Module`
  public. I think that's actually how we may want the public API to play
  out long-term as well, as it nicely allows for sub-borrows of just one
  class of item in wasm modules.

### DIFF
--- a/src/ir/mod.rs
+++ b/src/ir/mod.rs
@@ -18,6 +18,7 @@ use crate::ty::TypeId;
 use crate::ty::ValType;
 use id_arena::Id;
 use std::fmt;
+use std::mem;
 use walrus_derive::walrus_expr;
 
 /// The id of a local.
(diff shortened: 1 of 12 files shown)


src/ir/mod.rs
```diff
@@ -894,6 +903,24 @@ impl<'expr> Visit<'expr> for ExprId {
...
 
+impl VisitMut for ExprId {
+    fn visit_mut<V>(&mut self, visitor: &mut V)
+    where
+        V: VisitorMut,
+    {
+        // TODO: this is somewhat unfortunate and seems like it's susceptible to
+        // being bug-prone when we have a DAG of an IR. Is there perhaps a
+        // different pattern we should be using? Maybe remove methods from
+        // `VisitMut` or something like that?
```
fitzgen: At that point we will want to just have a graph traversal where nodes are `ExprId` and let folks do the deref + match manually I think.

alexcrichton: Ah good point, I'll add this to the comment here

PR #39, opened 2019-02-01
Ensure spec tests actually run, and fix some failures

alexcrichton:
More details in the commits!

### DIFF
--- a/src/module/functions/local_function/context.rs
+++ b/src/module/functions/local_function/context.rs
@@ -1,7 +1,7 @@
 //! Context needed when validating instructions and constructing our `Expr` IR.
 
 use crate::error::{ErrorKind, Result};
-use crate::ir::{Block, BlockId, BlockKind, ExprId};
+use crate::ir::{Block, BlockId, BlockKind, ExprId, Drop};
 use crate::module::functions::{FunctionId, LocalFunction};
 use crate::module::Module;
 use crate::parse::IndicesToIds;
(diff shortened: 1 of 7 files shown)


src/passes/used.rs
```diff
@@ -64,6 +65,32 @@ impl Used {
             stack.push_func(f);
         }
 
+        // Initialization of imported memories or imported tables is a
```
alexcrichton: Note that this is arguably a bug in wasm-gc as-is today!

walrus-tests/tests/spec-tests.rs
```diff
@@ -115,8 +115,21 @@ fn run(wast: &Path) -> Result<(), failure::Error> {
         .args(extra_args)
         .output()
         .context("executing `spectest-interp`")?;
+
+    // If the interpreter exits with success it may still have failed some
+    // tests. Check the output for `X/Y tests passed.` and make sure `X` equals
+    // `Y`.
     if output.status.success() {
```
fitzgen: Seems like maybe we should file a bug upstream to exit with code = number of failures or something.

src/module/functions/local_function/context.rs
```diff
@@ -143,8 +143,30 @@ impl<'a> FunctionContext<'a> {
...
         let expr = expr.into();
-        self.add_to_current_frame_block(expr);
-        impl_unreachable(&mut self.operands, &mut self.controls, expr);
+
+        let frame = self.controls.last_mut().unwrap();
+
+        // If we're not unreachable yet then we need to commit this expression.
+        // Note that we also need to commit any previous expressions on the
+        // stack because they may have side effects. For any lingering operands
+        // synthesize `Drop` nodes for them here.
```
fitzgen: Nice.

src/passes/used.rs
fitzgen: > If a module imports a table/memory and has initializers for it, then we should consider that import live and retained to ensure initializers happen when the module is instantiated.

Yeah, that was an existing bug for sure

walrus-tests/tests/spec-tests.rs
alexcrichton: https://github.com/WebAssembly/wabt/issues/1002

PR #47, opened 2019-02-11
Add an option to skip validation after parsing

alexcrichton:
We'll at least need this for parsing LLVM's output which currently has
atomic instructions but doesn't have shared memory!

### DIFF
--- a/src/module/mod.rs
+++ b/src/module/mod.rs
@@ -205,7 +205,9 @@ impl Module {
             .add_processed_by("walrus", env!("CARGO_PKG_VERSION"));
 
         // TODO: probably run this in a different location
-        crate::passes::validate::run(&ret)?;
+        if !ret.config.skip_strict_validate {
+            crate::passes::validate::run(&ret)?;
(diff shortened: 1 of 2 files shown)


src/module/mod.rs
```diff
@@ -205,7 +205,9 @@ impl Module {
             .add_processed_by("walrus", env!("CARGO_PKG_VERSION"));
 
         // TODO: probably run this in a different location
-        crate::passes::validate::run(&ret)?;
+        if !ret.config.skip_validate {
+            crate::passes::validate::run(&ret)?;
```
fitzgen: So we still do *most* validation, just not the stuff that happens after parsing? Seems like maybe we should rename this or at least clarify...

alexcrichton: Indeed yeah, although I think we'll long-term want to move all the type checking validation done during instructin parsing to the explicit validation pass if we can. In any case *some* validation happens when parsing still so I'll clarify the name here

PR #51, opened 2019-02-14
No dwarf

fitzgen:


src/module/mod.rs
```diff
@@ -256,9 +256,17 @@ impl Module {
         self.funcs.emit(&mut cx);
         self.data.emit(&mut cx);
 
-        emit_name_section(&mut cx);
+        if self.config.generate_names {
```
alexcrichton: Oh so the `generate_names` here was intended for debugging passes in walrus itself, so a lot of debug information is tied to this currently. I think we may want a separate flag (on by default whereas the `generate_names` flag is off by default) to emit the name section?

src/module/mod.rs
```diff
@@ -256,9 +256,17 @@ impl Module {
...
+            emit_name_section(&mut cx);
+        }
+
         self.producers.emit(&mut cx);
         for section in self.custom.iter() {
+            if !self.config.generate_dwarf && section.name.starts_with(".debug") {
```
alexcrichton: I wonder if this should also be skipped when parsing to avoid keeping the dwarf sections in memory?

PR #54, opened 2019-02-15
Expose a `FunctionBuilder` through `LocalFunction`

alexcrichton:
This should allow to continue to build expressions for a function even
after it exists

### DIFF
--- a/src/module/functions/local_function/mod.rs
+++ b/src/module/functions/local_function/mod.rs
@@ -13,9 +13,9 @@ use crate::ir::*;
 use crate::map::{IdHashMap, IdHashSet};
 use crate::parse::IndicesToIds;
 use crate::passes::Used;
-use crate::{FunctionId, Module, Result, TypeId, ValType, TableKind};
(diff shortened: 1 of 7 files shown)


src/module/functions/local_function/mod.rs
```diff
@@ -116,22 +115,34 @@ impl LocalFunction {
...
 
     /// Get the expression associated with the given id
     pub fn get_mut(&mut self, id: ExprId) -> &mut Expr {
-        &mut self.exprs[id]
+        &mut self.exprs.arena[id]
+    }
+
+    /// Get access to a `FunctionBuilder` to continue adding expressions to
+    /// this function.
+    pub fn builder(&self) -> &FunctionBuilder {
```
fitzgen: One can't use a shared ref to a function builder to add expressions though, right?

alexcrichton: True! I figured it'd be good to expose in terms of idioms although we can always remove it later

PR #57, opened 2019-02-15
Add a convenience for getting the function tables

alexcrichton:
If applicable modules typically have one function table when coming from
languages like C or Rust, so add a convenience function that returns the
function table if it's present.

### DIFF
--- a/src/module/tables.rs
+++ b/src/module/tables.rs
@@ -132,6 +132,31 @@ impl ModuleTables {
     pub fn iter(&self) -> impl Iterator<Item = &Table> {
         self.arena.iter().map(|p| p.1)
     }
+
+    /// Finds a unique function table in a module.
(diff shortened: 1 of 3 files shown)


src/module/tables.rs
```diff
@@ -132,6 +132,31 @@ impl ModuleTables {
...
+    ///
+    /// Modules produced by compilers like LLVM typically have one function
+    /// table for indirect function calls. This function will look for a single
+    /// function table inside this module, and return that if found. If no
+    /// function tables are present `None` will be returned
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if there are two function tables in this module
+    pub fn single_function_table(&self) -> Result<Option<TableId>> {
```
fitzgen: `main_function_table`?

PR #60, opened 2019-02-19
Add an explicit GC pass to walrus

alexcrichton:
Currently we implicity GC a module when emitting it, but this instead
adds an explicit GC step with the new `delete` functions that we have
now. This way consumers can explicitly choose when to `gc` and then act
on the decisions of the GC to emit different bindings, for example.

### DIFF
--- a/crates/tests/tests/spec-tests.rs
+++ b/crates/tests/tests/spec-tests.rs
@@ -48,23 +48,20 @@ fn run(wast: &Path) -> Result<(), failure::Error> {
 
     let contents = fs::read_to_string(&json).context("failed to read file")?;
     let test: Test = serde_json::from_str(&contents).context("failed to parse file")?;
-    let prev_len = test.commands.len();
-    let mut new_commands = Vec::new();
+    let mut files = Vec::new();
+
+    let mut config = walrus::ModuleConfig::new();
+    if extra_args.len() == 0 {
+        config.only_stable_features(true);
+    }
 
     for command in test.commands {
         let filename = match command.get("filename") {
             Some(name) => name.as_str().unwrap().to_string(),
-            None => {
-                new_commands.push(command);
-                continue;
-            }
+            None => continue,
         };
         let line = &command["line"];
         let path = tempdir.path().join(filename);
-        let mut config = walrus::ModuleConfig::new();
-        if extra_args.len() == 0 {
-            config.only_stable_features(true);
-        }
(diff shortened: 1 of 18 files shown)


crates/tests/tests/spec-tests.rs
```diff
@@ -120,27 +94,42 @@ fn run(wast: &Path) -> Result<(), failure::Error> {
...
+    for (cmd, file) in files.iter() {
+        let wasm = fs::read(file)?;
+        let mut module = config.parse(&wasm)?;
+
+        // Tests which assert that they're not linkable tend to not work with
+        // the gc pass because it removes things which would cause a module to
+        // become unlinkable. This doesn't matter too much in the real world
+        // (hopefully), so just don't gc assert_unlinkable modules.
+        if cmd != "assert_unlinkable" {
+            walrus::passes::gc::run(&mut module);
```
fitzgen: this is a lot cleaner than the old fix :+1:

src/module/tables.rs
```diff
@@ -200,28 +200,17 @@ impl Module {
...
-            cx.used.tables.contains(&table.id) && table.import.is_none()
-        };
-
-        let tables = self
-            .arena
-            .iter()
-            .filter(|(_id, table)| emitted(cx, table))
-            .count();
-
+        // Skip imported tables because those are emitted elsewhere
```
fitzgen:
```suggestion
        // Skip imported tables because those are emitted in the import section.
```

src/module/types.rs
```diff
@@ -29,6 +29,11 @@ impl ModuleTypes {
         self.arena.iter().map(|(_, f)| f)
     }
 
+    /// Delete a type from this module
+    pub fn delete(&mut self, ty: TypeId) {
```
fitzgen: Worth adding a comment about references to the deleted type, like the other `delete` methods.

alexcrichton:
Currently I put `wasm-gc` sort of on-ice as in theory nothing in our toolchain requires it, but it's definitely trivial to implement with this pass now :)

fitzgen:
> Currently I put `wasm-gc` sort of on-ice as in theory nothing in our toolchain requires it, but it's definitely trivial to implement with this pass now :)

Yeah, but I suspect there are random folks using it who aren't using the rest of our toolchain. Maybe their needs aren't the highest priority though, and it isn't like the tool is currently broken...

PR #65, opened 2019-03-04
WIP: add name method to FunctionBuilder

### LINKED ISSUE #61: Add a `name` method to `FunctionBuilder`
This would set the function's name for use with the name section.

Right now the name lives in the `Function` itself, we would want to move that into the builder, and then expose a method on the builder for setting the name, and a getter on `Function` for getting its name.

See `src/function_builder.rs` and `src/functions/mod.rs`.

fgimenez:
Closes #61

### DIFF
--- a/src/function_builder.rs
+++ b/src/function_builder.rs
@@ -9,6 +9,9 @@ use std::ops::{Deref, DerefMut, Drop};
 #[derive(Default, Debug)]
 pub struct FunctionBuilder {
     pub(crate) arena: TombstoneArena<Expr>,
+
+    /// An optional name associated with this function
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks @fgimenez! I think the current way it's set up we may not actually be able to remove the `name` field of `Function`. Only a `LocalFunction` stores a `FunctionBuilder`, but non-locally-defined functions otherwise don't have storage space for their name. 

This is probably something we'll improve in the future though!

fgimenez:
sure, thanks @fitzgen and @alexcrichton! on it, will ping you when ready.

PR #74, opened 2019-05-13
Update encoding of `table.grow`

alexcrichton:
Looks like the order of its operands have swapped since this was last
tested!


alexcrichton:
Currently tests are failing due to https://github.com/WebAssembly/wabt/pull/1079 not being merged yet, but I suspect that won't take too long to land.

PR #76, opened 2019-05-17
Custom sections API

### LINKED ISSUE #75: Provide APIs for working with custom sections
### Motivation

There are many custom sections that are important for different use cases, eg webidl bindings custom section or the linking custom sections. People should be able to read, write, and manipulate custom sections that `walrus` itself doesn't know about.

### Proposed Solution

```rust
trait CustomSection: Any + Send {
    fn name(&self) -> &'static str;

    fn emit(&self, cx: &mut EmitContext);
}

/// One of these will be in `Module`.
pub struct ModuleCustomSections { ... }

impl ModuleCustomSections {
    /// Take the raw bytes of an unparsed custom section out of this module, if present.
    pub fn take_raw(&mut self, name: &str) -> Option<Vec<u8>>;

    /// Insert the given custom section into the module.
    pub fn add<T>(&mut self, custom_section: T) -> CustomSectionId<T>
        where T: CustomSection;

    /// Gets the given custom section that was previously added or panics.
    pub fn get<T>(&self, id: CustomSectionId<T>) -> &T
        where T: CustomSection;
    pub fn get_mut<T>(&mut self, id: CustomSectionId<T>) -> &mut T
        where T: CustomSection;

    /// Iterate over custom sections.
    pub fn iter(&self) -> impl Iterator<&dyn CustomSection>;
    pub fn iter_mut(&self) -> impl Iterator<&mut dyn CustomSection>;
}
```

This way, we don't have to specify an API for parsing, instead we let consumers take the raw bytes, parse however they want, and optionally add custom sections back into the module for emitting (which we do need to know how to emit).

### Alternatives

We could make parsing part of the `CustomSection` trait. This would make it more complex and less flexible, however.

We could avoid any trait object at all, and just take and receive `Vec<u8>` for custom sections. This has the advantage of not needing to expose the encoder machinery publicly. However, this forces users to have somewhere else to stash the parsed custom sections while manipulating a module, or to keep re-parsing and re-serializing them.

------------

Thoughts @alexcrichton?

fitzgen:
Fixes #75

### DIFF
--- a/crates/tests/tests/custom_sections.rs
+++ b/crates/tests/tests/custom_sections.rs
@@ -1,0 +1,33 @@
+//! Tests for working with custom sections that `walrus` doesn't know about.
+
+use std::borrow::Cow;
+use walrus::{CustomSection, Module};
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+struct HelloCustomSection(String);
+
+impl CustomSection for HelloCustomSection {
+    fn name(&self) -> &str {
+        "hello"
+    }
+
+    fn data(&self) -> Cow<[u8]> {
+        let data = format!("Hello, {}!", self.0);
+        data.into_bytes().into()
+    }
+}
+
+#[test]
+fn round_trip_unkown_custom_sections() {
+    let mut module = Module::default();
+
+    let world = HelloCustomSection("World".into());
+    let world_id = module.customs.add(world.clone());
+    assert_eq!(module.customs.get(world_id), &world);
+
+    let wasm = module.emit_wasm().unwrap();
+    let mut module = Module::from_buffer(&wasm).unwrap();
+
+    let world_round_tripped = module.customs.take_raw("hello").unwrap();
+    assert_eq!(world_round_tripped.data(), world.data());
+}
--- a/crates/tests/tests/custom_sections.rs
+++ b/crates/tests/tests/custom_sections.rs
@@ -1,10 +1,21 @@
 //! Tests for working with custom sections that `walrus` doesn't know about.
 
 use std::borrow::Cow;
-use walrus::{CustomSection, Module};
+use walrus::{CustomSection, Module, ModuleConfig};
 
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 struct HelloCustomSection(String);
+
+impl HelloCustomSection {
+    fn parse(data: &[u8]) -> Option<Self> {
+        let data = std::str::from_utf8(data).ok()?;
+        if !data.starts_with("Hello, ") || !data.ends_with("!") {
+            return None;
+        }
+        let who = data["Hello, ".len()..data.len() - 1].to_string();
+        Some(HelloCustomSection(who))
+    }
+}
 
 impl CustomSection for HelloCustomSection {
     fn name(&self) -> &str {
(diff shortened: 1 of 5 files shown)


src/module/custom.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /// Get a shared reference to a custom section that is in this
+    /// `ModuleCustomSections`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the custom section associated with the given `id` has been
+    /// deleted or taken.
+    pub fn get<T>(&self, id: TypedCustomSectionId<T>) -> &T
```
alexcrichton: Following the precedent of std it may be best to remove `try_*` and have this family of functions return `Option`?

src/module/custom.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn delete<I>(&mut self, id: I)
+    where
+        I: Into<CustomSectionId>,
+    {
+        let id = id.into().0;
+        self.arena.delete(id);
+    }
+
+    /// Take a raw, unparsed custom section out of this module.
+    pub fn take_raw(&mut self, name: &str) -> Option<RawCustomSection> {
```
alexcrichton: `remove_raw` perhaps?

src/module/custom.rs
```diff
@@ -0,0 +1,279 @@
...
+            .arena
+            .alloc(Some(Box::new(custom_section) as Box<dyn CustomSectionAny>));
+        TypedCustomSectionId {
+            id: CustomSectionId(id),
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Remove a custom section from the module.
+    pub fn delete<I>(&mut self, id: I)
```
alexcrichton: Perhaps `remove` to align with libstd collections?

src/module/custom.rs
```diff
@@ -0,0 +1,279 @@
...
+///   back into the module, so `walrus` can emit the processed/updated version
+///   of the custom section.
+#[derive(Debug, Default)]
+pub struct ModuleCustomSections {
+    arena: TombstoneArena<Option<Box<dyn CustomSectionAny>>>,
+}
+
+impl ModuleCustomSections {
+    /// Add a new custom section to the module.
+    pub fn add<T>(&mut self, custom_section: T) -> TypedCustomSectionId<T>
```
alexcrichton: Perhaps `insert` / `push` to align with libstd collections?

src/module/custom.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /// Get an untyped, shared reference to a custom section that is in this
+    /// `ModuleCustomSections`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the custom section associated with the given `id` has been
+    /// deleted or taken.
+    pub fn get_untyped(&self, id: CustomSectionId) -> &dyn CustomSection {
```
alexcrichton: So this is definitely a nitpick and not at all necessary for us to handle now. But I feel like this wants something like:

```rust
pub fn get<T>(&self, id: T) -> Option<&T::CustomSection>;
```

where you can `get(CustomSectionId)` to get `Option<&dyn CustomSection>` or you can `get(TypedSectionId<T>) -> Option<&T>`.

Basically just moving the typed/untyped distinction to the ID, and otherwise only having APIs that take ids and not some for typed ids and others for untyped ids.

crates/tests/tests/custom_sections.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    let world = HelloCustomSection("World".into());
+    let world_id = module.customs.add(world.clone());
+    assert_eq!(module.customs.get(world_id), &world);
+
+    let wasm = module.emit_wasm().unwrap();
+    let mut module = Module::from_buffer(&wasm).unwrap();
+
+    let world_round_tripped = module.customs.take_raw("hello").unwrap();
+    assert_eq!(world_round_tripped.data(), world.data());
```
alexcrichton: Could this also exercise removing the raw section, parsing it, and then inserting it back into the module? (just running thorugh the APIs to discover accidental hidden panics

src/module/custom.rs
fitzgen: Our other APIs (eg for tables and functions, etc) use `delete`, so I think we should leave it in this PR, and maybe file a good-first-issue for switching all of them en mass.

src/module/custom.rs
fitzgen: Similar story to `delete` here: all our existing APIs use some variation of `add`, so I'd prefer to align with that.

src/module/custom.rs
fitzgen: Nice idea.

crates/tests/tests/custom_sections.rs
fitzgen: For sure.

src/module/custom.rs
fitzgen: And then all the things that are currently `get` should be `Index` impls?

alexcrichton: In theory yeah, although I'd be fine leaving the `Index` impls for later.

PR #81, opened 2019-05-24
More enhancements for custom sections

alexcrichton:
* Make sure to export all types
* Return the custom section value on `delete`
* Add convenience methods for accessing sections based on type rather than ID

### DIFF
--- a/src/module/custom.rs
+++ b/src/module/custom.rs
@@ -276,7 +276,7 @@ impl ModuleCustomSections {
     /// Remove a custom section from the module.
     pub fn delete<I>(&mut self, id: I) -> Option<Box<I::CustomSection>>
     where
-        I: CustomSectionId
+        I: CustomSectionId,
     {
         let id = id.into_inner_id();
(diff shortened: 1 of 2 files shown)


src/module/custom.rs
```diff
@@ -345,4 +357,42 @@ impl ModuleCustomSections {
             }
         })
     }
+
+    /// Remove a custom section (by type) from the module.
+    ///
+    /// If there are multiple custom sections of the type `T` only the first one
+    /// is removed.
```
fitzgen: "first" is a bit loosely defined, since the arena's items can't really be reordered. ah well.

src/module/mod.rs
```diff
@@ -19,6 +19,7 @@ use crate::encode::Encoder;
 use crate::error::Result;
 pub use crate::module::custom::{
     CustomSection, CustomSectionId, ModuleCustomSections, RawCustomSection, TypedCustomSectionId,
+    UntypedCustomSectionId,
```
fitzgen: woops 😅

PR #87, opened 2019-06-04
Stuff needed for webidl bindings

fitzgen:


src/module/functions/mod.rs
```diff
@@ -182,6 +182,20 @@ impl ModuleFunctions {
         &mut self.arena[id]
     }
 
+    /// Get a function ID by its name.
+    ///
+    /// Note that function names are *not* guaranteed to be unique. This will
+    /// return the first function in the module with the given name.
```
alexcrichton: It might be good to mention here that the filtering is done by the name subsection, not by the export name

src/module/types.rs
```diff
@@ -23,6 +23,17 @@ impl ModuleTypes {
         &mut self.arena[id]
     }
 
+    /// Get a type ID by its name.
```
alexcrichton: Perhaps worth mentioning that this is only intended for in-memory modifications and by default this always returns `None` for any newly-deserialized module?

fitzgen:
> Could you expand more on the need to preserve `IndicesToIds`? I was hoping we could continue forgetting about that and deleting it after parsing but I suspect you've got a good reason for preserving it!

Sure thing -- when parsing the text format for webidl bindings, types and functions can be referenced by either name or index, same as in the normal wasm text format. So when we get around to parsing the webidl text, we need access to this stuff to translate it into our own webidl-thing IDs.

If we required custom sections to be pre-registered somehow, so that we parse them all at the same time as we parse the rest of the non-custom sections, then we could throw the `IndicesToIds` away afterwards.

Since the thing is `pub`, one can always replace it with the default, non-heap-allocating instance as well (and we could provide a helper for that too, if we wanted).

alexcrichton:
Ok cool that makes sense. I think I'd idealistically prefer to have a system where the `IndicesToIds` type is explicitly only available during parsing, for example we'd have like a "parsing context" where you could parse out a module and then use that context to further parse custom sections. (or the idea of you preregister with a `ModuleConfig` the custom sections you'd like to parse).

Practically that'd all take far too long and wouldn't have a ton of benefit. In the meantime mind updating the documentation on the `indices_to_ids` field to indicate that it's primarily only intended for parsing custom sections and/or anything "during the parsing phase"? It also will be empty for modules synthesized in memory (etc)

alexcrichton:
r=me with the comment update is also what I meant to say

fitzgen:
> Ok cool that makes sense. I think I'd idealistically prefer to have a system where...

100% agree.

I wonder if we could optionally return the `IndicesToIds` along with the module in some `Module::parse_with_all_the_extra_shit` method...

alexcrichton:
Also sounds plausible to me! I'm fine with w/e strategy, can always revisit later

alexcrichton:
Nice.

PR #91, opened 2019-06-07
Add unwrap_* methods to FunctionKind

### LINKED ISSUE #82: Add `unwrap_*` methods to `FunctionKind` to unwrap each variant of the enum
Looking at https://github.com/rustwasm/wasm-bindgen/pull/1566 and seeing a bunch of stuff that could get cleaned up if we had such a method.

Atul9:
closes #82

### DIFF
--- a/src/module/functions/mod.rs
+++ b/src/module/functions/mod.rs
@@ -111,6 +111,15 @@ pub enum FunctionKind {
 }
 
 impl FunctionKind {
+    /// Get the underlying `FunctionKind::Import` or panic if this is not an
+    /// import function
+    pub fn unwrap_import(&self) -> &ImportedFunction {
+        match *self {
+            FunctionKind::Import(ref import) => import,
+            _ => panic!("not an import function"),
+        }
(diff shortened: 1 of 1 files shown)


src/module/functions/mod.rs
```diff
@@ -119,6 +128,15 @@ impl FunctionKind {
             _ => panic!("not a local function"),
         }
     }
+
+    /// Get the underlying `FunctionKind::Uninitialized` or panic if this is
+    /// not an uninitialized function
+    pub fn unwrap_uninitialized(&self) -> &TypeId {
```
fitzgen: Hm, I actually think we shouldn't expose this method, since the uninitialized variant is sort of semi-private, and you can't really do anything with it anyways.

Let's remove this method from the PR. Sorry for not catching this earlier!

Atul9: @fitzgen done. I have updated the PR.

alexcrichton:
Thanks!

PR #92, opened 2019-06-08
Return ImportId for `Module::add_import_*

### LINKED ISSUE #86: Return `ImportId` from `Module::add_import_*`
We have a number of functions on `Module` which allow adding an imported item which registers it with both the imports section as well as the appropriate section the item goes into. Currently only the target item's ID is returned, but we should also return the `ImportId` in addition to the item's actual ID.

c410-f3r:
Fixes #86.


alexcrichton:
Perfect!

PR #93, opened 2019-06-08
 Fix docs inside quote! 

### LINKED ISSUE #85: Macro-generated doc comments have "#name" instead of the interpolated name
For example: https://docs.rs/walrus/0.7.0/walrus/ir/enum.Expr.html#method.unwrap_call

We need to build the doc string outside of the `quote!` and then do `#[doc = #doc_comment]` rather than writing `/// ...` inside the `quote!`.

c410-f3r:
Fixes #85.


alexcrichton:
Nice!

PR #97, opened 2019-06-29
Correctly set block type in FunctionBuilder.loop_

OneSadCookie:
simple typo :)

Are there tests for this hiding somewhere that I should have updated?


alexcrichton:
Thanks! The builder here is light on tests but it's ok that there isn't much to uodate

PR #108, opened 2019-07-24
Fix `wasm-opt -ttf` fuzzing

fitzgen:
`wasm-opt -ttf` will assert that it produces valid wasm modules, but it seems that it can still generate sign extension instructions even with `--disable-sign-ext`, which causes validation to fail, and ultimately the process to exit unsuccessfully. Instead, we just skip and try again when `wasm-opt` fails, and do the same for if `wat2wasm` cannot interpret what `wasm-opt` generates (which was why we previously disabled the sign extension instructions).

See also https://github.com/WebAssembly/binaryen/issues/2257

### DIFF
--- a/crates/fuzz/src/lib.rs
+++ b/crates/fuzz/src/lib.rs
@@ -374,16 +374,19 @@ impl TestCaseGenerator for WasmOptTtf {
             let input_tmp = tempfile::NamedTempFile::new().unwrap();
             fs::write(input_tmp.path(), input).unwrap();
 
-            let wat = walrus_tests_utils::wasm_opt(
-                input_tmp.path(),
-                vec![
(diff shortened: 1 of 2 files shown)


fitzgen:
Note that we were incorrectly requiring that `data.drop` and `memory.init` operands be passive data segments. These are actually allowed during validation but result in runtime errors. See the commit message for details.

This latest commit should have us green again.

alexcrichton:
👍

alexcrichton:
Aww while CI passed here it [failed on the merge](https://travis-ci.org/rustwasm/walrus/jobs/563354538) :(

fitzgen:
I'll look into it some more.

PR #111, opened 2019-07-29
Make `rayon` an optional dependency

alexcrichton:
This commit makes `rayon` an optional dependency because it is known to
not work on targets such as WebAssembly by default. Additionally it's
not necessarily guaranteed that all users of `walrus` want its parallel
processing capabilties, so this may help it make a lighter dependency in
those situations!

The main use case for me though was I was playing around with some wasm
tooling in Rust compiled to wasm but `walrus` didn't work out-of-the-box
in that context due to its usage of parallel iterators.

### DIFF
--- a/src/module/functions/mod.rs
+++ b/src/module/functions/mod.rs
@@ -13,8 +13,10 @@ use crate::tombstone_arena::{Id, Tombstone, TombstoneArena};
 use crate::ty::TypeId;
 use crate::ty::ValType;
 use failure::bail;
-use rayon::prelude::*;
 use std::cmp;
+
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
 
 pub use self::local_function::LocalFunction;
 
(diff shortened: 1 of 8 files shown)


alexcrichton:
@fitzgen I'm curious to hear your thoughts on this in terms of maintainability, it's not too bad here but it is more unfortunate than I would have otherwise hoped for.

src/module/functions/mod.rs
```diff
@@ -219,6 +224,9 @@ impl ModuleFunctions {
     }
 
     /// Get a parallel iterator of this module's local functions
+    ///
+    /// Requires the `parallel` feature of this crate to be enabled.
+    #[cfg(feature = "parallel")]
     pub fn par_iter_local(&self) -> impl ParallelIterator<Item = (FunctionId, &LocalFunction)> {
```
fitzgen: What if instead of `cfg`ing this method into existence or not, if the parallel feature wasn't enabled, then we returned a normal iterator? I think this would have less of an impact on various call sites, and should Just Work, right? Urgh well I guess since this is a public function, and enabling the feature changes the signature, that's a no go :-/

fitzgen:
> @fitzgen I'm curious to hear your thoughts on this in terms of maintainability, it's not too bad here but it is more unfortunate than I would have otherwise hoped for.

Yeah, it isn't too bad, but a little more invasive than I'd hoped too.

alexcrichton:
Mk, I'm hoping that we're hit with a bolt of inspiration at some point to make this a bit nicer :)

PR #114, opened 2019-08-01
List IR

fitzgen:
Still need to re-add graphviz dot support, so not quite ready to land yet. But the most important bits are already here, so I figured you might want to take a look now, @alexcrichton.


fitzgen:
Take a look at the `build-wasm-from-scratch.rs` example to get an idea of how builders have changed: https://github.com/fitzgen/walrus/blob/list-ir/examples/build-wasm-from-scratch.rs#L52-L97

Visitors are really not that different. Although I'd also like to replace them with something that isn't recursive.

alexcrichton:
This looks great to me, thanks for taking this on @fitzgen! I think we'll probably want to flesh out the non-recursive visitation before this lands though to ensure we've got a working strategy for it.

Also would you be up for prototyping a patch to `wasm-bindgen` to use this `walrus` to double-check that it still works for `wasm-bindgen`?

fitzgen:
Yeah for sure

fitzgen:
@alexcrichton ok, I added some non-recursive traversals. See the latest commit. Still working on re-implementing graphviz support.

Issue #113, opened 2019-08-01
Add getters on `ModuleExports` for getting an Export via the exported thing's id


fitzgen:
For example, getting the `Export` that is an exported function (ie has `ExportKind::Function`) via a function id.

```rust
impl ModuleExports {
    pub fn get_exported_func(&self, f: FunctionId) -> Option<&Export> { ... }
    pub fn get_exported_table(&self, f: TableId) -> Option<&Export> { ... }
    pub fn get_exported_memory(&self, f: TableId) -> Option<&Export> { ... }
    pub fn get_exported_global(&self, f: TableId) -> Option<&Export> { ... }
}
```

lshlyapnikov:
@fitzgen how can I convert `FunctionId` and `TableId` to `ExportId`?
We have:
```
pub struct ModuleExports {
    /// The arena containing this module's exports.
    arena: TombstoneArena<Export>,
}
```
so, in order to fetch an export from the arena, I need an `ExportId`. Unless you want to iterate or have separate arenas for functions and tables.

lshlyapnikov:
Or maybe we can accumulate the `FunctionId -> ExportId` and `TableId -> ExportId` mappings as part of the `add`:
```
    /// Add a new export to this module
    pub fn add(&mut self, name: &str, item: impl Into<ExportItem>) -> ExportId {
        self.arena.alloc_with_id(|id| Export {
            id,
            name: name.to_string(),
            item: item.into(),
        })
    }
```
I think I will implement this. Let me know if there is any other/better solution.

lshlyapnikov:
@fitzgen I also believe you meant:
```
impl ModuleExports {
    pub fn get_exported_func(&self, f: FunctionId) -> Option<&Export> { ... }
    pub fn get_exported_table(&self, f: TableId) -> Option<&Export> { ... }
    pub fn get_exported_memory(&self, f: MemoryId) -> Option<&Export> { ... }
    pub fn get_exported_global(&self, f: GlobalId) -> Option<&Export> { ... }
}
```
the type of `f` argument for the last two functions.

fitzgen:
> @fitzgen I also believe you meant:

Yes -- good eye!

> I think I will implement this. Let me know if there is any other/better solution.

Sounds great!

lshlyapnikov:
@fitzgen not sure if you got notified... please have a look at this comment: https://github.com/rustwasm/walrus/pull/115#issuecomment-521487860

[...]

PR #115, opened 2019-08-10
Adding ModuleExports getters for FunctionId, TableId, MemoryId and GlobalId

lshlyapnikov:
See #113 
## TODO:
- [x] Make sure updates via `ModuleExports::iter_mut()` get reflected in the arena (an issue with IdHashMap's not being updated)
- [x] Make sure there are no other ways to remove exports except `ModuleExports::delete`
- [x] Add `FunctionId` test case
- [x] Add `TableId` test case
- [x] Add `MemoryId` test case
- [x] Add `GlobalId` test case

### DIFF
--- a/src/module/exports.rs
+++ b/src/module/exports.rs
@@ -240,34 +240,113 @@ impl From<TableId> for ExportItem {
 mod tests {
     use super::*;
     use crate::{FunctionBuilder, Module};
+    use id_arena::Arena;
 
     #[test]
     fn get_exported_func() {
         let mut module = Module::default();
         let mut builder = FunctionBuilder::new(&mut module.types, &[], &[]);
         builder.func_body().i32_const(1234).drop();
-        let function_id = builder.finish(vec![], &mut module.funcs);
-        module.exports.add("dummy", function_id);
+        let id: FunctionId = builder.finish(vec![], &mut module.funcs);
+        module.exports.add("dummy", id);
 
-        let actual: Option<&Export> = module.exports.get_exported_func(function_id);
+        let actual: Option<&Export> = module.exports.get_exported_func(id);
 
         let export: &Export = actual.expect("Expected Some(Export) got None");
         assert_eq!(export.name, "dummy");
         match export.item {
-            ExportItem::Function(f) => assert_eq!(f, function_id),
-            _ => panic!("Expected a Function"),
+            ExportItem::Function(f) => assert_eq!(f, id),
+            _ => panic!("Expected a Function variant"),
         }
     }
 
     #[test]
     fn get_exported_func_should_return_none_for_unknown_function_id() {
-        let mut module = Module::default();
-        let mut builder = FunctionBuilder::new(&mut module.types, &[], &[]);
-        builder.func_body().i32_const(1234).drop();
-        let function_id = builder.finish(vec![], &mut module.funcs);
+        let module = Module::default();
+        let arena: Arena<Function> = Arena::new();
+        let id: FunctionId = arena.next_id();
 
-        let actual: Option<&Export> = module.exports.get_exported_func(function_id);
+        let actual: Option<&Export> = module.exports.get_exported_func(id);
 
         assert!(actual.is_none());
     }
+
+    #[test]
+    fn get_exported_table() {
+        let mut module = Module::default();
+        let arena: Arena<Table> = Arena::new();
+        let id: TableId = arena.next_id();
+        module.exports.add("dummy", id);
+
+        let actual: Option<&Export> = module.exports.get_exported_table(id);
+
+        let export: &Export = actual.expect("Expected Some(Export) got None");
+        assert_eq!(export.name, "dummy");
+        match export.item {
+            ExportItem::Table(f) => assert_eq!(f, id),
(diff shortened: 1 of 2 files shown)


src/module/exports.rs
```diff
@@ -49,6 +51,10 @@ pub enum ExportItem {
 pub struct ModuleExports {
     /// The arena containing this module's exports.
     arena: TombstoneArena<Export>,
+    fn_id_to_export_id: HashMap<FunctionId, ExportId>,
+    tbl_id_to_export_id: HashMap<TableId, ExportId>,
+    mem_id_to_export_id: HashMap<MemoryId, ExportId>,
+    global_id_to_export_id: HashMap<GlobalId, ExportId>,
```
fitzgen: Instead of using the default `HashMap` with the default hasher, we have a faster version for specifically working with ids:

https://github.com/rustwasm/walrus/blob/42d3b6116568798309797dad4ede9880835dbecc/src/map.rs#L18-L19

src/module/exports.rs
```diff
@@ -62,8 +68,17 @@ impl ModuleExports {
         &mut self.arena[id]
     }
 
+    fn get_safe(&self, id: ExportId) -> Option<&Export> {
+        self.arena.get(id)
+    }
+
+    fn get_mut_safe(&mut self, id: ExportId) -> Option<&mut Export> {
```
fitzgen: "safe" is a misnomer here. For now, I think we should just access the arenas directly in the public methods below.

Longer term, we should make the regular `get` variants return options and not panic.

lshlyapnikov: Count me in for this refactoring. I wish it was a mono-repo :)

lshlyapnikov:
@fitzgen quick question: can we remove `ModuleExports::iter_mut() function`? Is it being used?

The issue with it, it allows making modifications to the exports without updating the `IdHasMap`s:
```
    fn_id_to_export_id: IdHashMap<Function, ExportId>,
    tbl_id_to_export_id: IdHashMap<Table, ExportId>,
    mem_id_to_export_id: IdHashMap<Memory, ExportId>,
    global_id_to_export_id: IdHashMap<Global, ExportId>,
```
here is an example:
```
let export_id: ExportId = module.exports.add("dummy", fn_id0);
module
    .exports
    .iter_mut()
    .for_each(|x: &mut Export| x.item = ExportItem::Function(fn_id1));
```
If we really need to keep the mutable iterator, I will have to remove the `IdHashMap`s and implement `get_exported_func`, `get_exported_table`, `get_exported_memory` and `get_exported_global` using `iterator().find()` which means the complexity will change from `O(1)` to `O(n)`.

Let me know what you prefer. I'd say let's just kill the `ModuleExports::iter_mut()` and if any something needs to be removed from the `ModuleExports` users might create a new instance without specific exports.

Let me know what you think.

fitzgen:
Good catch! It is being used by `wasm-bindgen`, although `wasm-bindgen` will skip non-function exports, which tells me that the use case is really "iterate over all exported functions" or "iterate over all exported memories" etc. If we only ever hand out `&mut Function`s instead of `&mut Export` or `&mut ExportItem` then user code can't change the item.

What do you think of switching this to `iter_funcs_mut()` etc?

lshlyapnikov:
> What do you think of switching this to `iter_funcs_mut()` etc?

Yeah we can do this... but it is just I don't see a good reason to allow users changing `Export.item` values. How about making this field private? `Export.id` is already private, we can make `item` private and we are good, the end users will not be able to change FunctionId, MemoryId, etc... So the ID mappings won't change:
https://github.com/rustwasm/walrus/blob/8cebc2ce2029bf02400ddac0db5bac89927b1feb/src/module/exports.rs#L15-L21
The only thing they will be allowed to change is the `Export.name`.

lshlyapnikov:
> If we only ever hand out `&mut Function`s instead of `&mut Export` or `&mut ExportItem` then user code can't change the item.

thinking about this... if they can't change `FunctionId` in `ExportItem::Function`, why do we need to provide a mutable iterator? :)

fitzgen:
> if they can't change `FunctionId` in `ExportItem::Function`, why do we need to provide a mutable iterator? :)

Sorry, I meant `&mut FunctionId`!

The use case we want to support is essentially generating thin wrappers around an exported function and then swapping the export to the wrapper function and making the wrapped function internal to the module.

lshlyapnikov:
> Sorry, I meant `&mut FunctionId`!
> 
Ok, makes sense... BTW how many exports do we expect? tens, hundreds, thousands? Or we don't know and this can differ per use case. I am just thinking if we can leave with O(N), assuming N is reasonably small. This would simplify the entire solution. We would not need to keep the ID mappings.

fitzgen:
The usual case is probably on the order of tens to hundreds, but there is probably a pretty long tail that is in the thousands.

I think either way will probably be fine, honestly.

lshlyapnikov:
@fitzgen Please have a look, it is ready for review. I have removed the `IdHashMap`s, this simplified a few things.

fitzgen:
Looks great! Thanks @lshlyapnikov :)

PR #124, opened 2019-08-29
Multi value! And even more fuzzing!

### LINKED ISSUE #77: Support multi-value proposal
### Motivation

The [multi-value proposal](https://github.com/WebAssembly/multi-value/blob/master/proposals/multi-value/Overview.md) is stage 3, and is getting pretty mature. We should support it.

Additionally, they will end up pretty heavily used by webidl-bindings.

### Proposed Solution

There are some tricky bits here. A single expression can produce multiple values now, so a single expression AST node can be the input to multiple other AST nodes, so we go from somethign that can be encoded as a tree to something that requires a DAG. The DAG must still be stack-y, however, so there are some constraints on where edges can be.

I propose that

* We annotate edges between expressions with an index of which value is being used. So if an expression E returns 3 values v<sub>0</sub>, v<sub>1</sub>, v<sub>2</sub>, and the second value v<sub>1</sub> is an input operand to expression F, we get an edge like `F ---1---> E`.

* We add a legalization pass that detects non-stack-y edges in the DAG and spills values to new locals as needed.

* By the time we emit, we assert that we don't ever see non-stack-y edges, and panic if we do.

### Alternatives

The main alternatives I can think of are:

1. Abandon our AST-based design and switch to basic blocks and CFGs. Then all these things are just SSA values and there isn't really much different between one return value and many. But now we have to implement relooper or some other method of structural analysis/recovery.

2. Implement hacks to special case multi return values in our IR and always immediately spill them into locals.

I think (1) is more work than this proposal and doesn't give us any clear gains over this proposal.

I think (2) is 🤮 and will lead to tech debt.

fitzgen:
fixes #77


fitzgen:
It looks like some change here (probably related to blocks having types or making the types shared) is making the types come out in non-deterministic orders. I'm going to solve this with an explicit sorting step for emitting types.

fuzz/fuzz_targets/raw.rs
```diff
@@ -0,0 +1,8 @@
+#![no_main]
+
+#[macro_use]
+extern crate libfuzzer_sys;
+
+fuzz_target!(|data: &[u8]| {
+    let _ = walrus::Module::from_buffer(data);
```
alexcrichton: Want to test the serialization here as well and if it succeeds at deserializing we immediately serialize it back out as well?

publish.sh
```diff
@@ -11,5 +11,8 @@ set -eux
 cd "$(dirname "$0")/crates/macro"
 cargo publish
 
+# Let crates.io's index notice that we published the macro.
+sleep 10
```
alexcrichton: FWIW when I do this locally I just use `--no-verify` to work around this

src/module/functions/local_function/mod.rs
```diff
@@ -247,10 +247,38 @@ impl LocalFunction {
...
+        }
+    }
+}
+
+fn block_param_tys(
+    ctx: &ValidationContext,
+    ty: wasmparser::TypeOrFuncType,
+) -> Result<Arc<[ValType]>> {
+    match ty {
+        wasmparser::TypeOrFuncType::Type(_) => Ok([][..].into()),
```
alexcrichton: I think this `.into()` can be somewhat expensive, so I wonder if a cached version of an empty list should be stored either in `ValidationContext` or in `module.types`?

src/module/types.rs
```diff
@@ -6,22 +6,66 @@ use crate::error::Result;
...
 use crate::parse::IndicesToIds;
 use crate::ty::{Type, TypeId, ValType};
+use std::sync::{Arc, Mutex};
 
 /// The set of de-duplicated types within a module.
+///
 #[derive(Debug, Default)]
 pub struct ModuleTypes {
-    arena: ArenaSet<Type>,
+    arena: Mutex<ArenaSet<Type>>,
```
alexcrichton: Did this basically just get way too unwieldy unless we had interior mutability?

src/module/types.rs
```diff
@@ -6,22 +6,66 @@ use crate::error::Result;
...
-    pub fn get_mut(&mut self, id: TypeId) -> &mut Type {
-        &mut self.arena[id]
+    /// Get a type associated with an ID mutably.
+    pub fn with_mut<T>(&self, id: TypeId, f: impl FnOnce(&mut Type) -> T) -> T {
+        let mut arena = self.arena.lock().unwrap();
+        f(&mut arena[id])
+    }
+
+    /// Iterate over each of this module's types.
+    pub fn each(&self, mut f: impl FnMut(&Type)) {
```
alexcrichton: This is sort of a bummer to have, but I wonder if we could have a `lock()` method or something like that which then provides iteration on the returned type?

src/module/types.rs
fitzgen: It became unweildy when blocks were storing their param and result types directly, and then doing a get-or-create for those param/result types during emit time. But I evolved the design since then, and `InstrSeq`s hold a `TypeId` directly now. They still do a get-or-create during parsing, which happens (can happen) in parallel for each function, so the synchronization is still required. However, it would be easy to switch that into a fallible `try-and-get-the-existing-type-with-these-params-and-results` method that returns an error instead of inserting when it doesn't already exist. This should be fine since the blocks reference an existing type in the binary format.

I'll try it out!

src/module/functions/local_function/mod.rs
fitzgen: Sure thing.

publish.sh
fitzgen: That works too, I suppose :-p

fuzz/fuzz_targets/raw.rs
fitzgen: Great idea!

src/module/types.rs
alexcrichton: Ah ok, seems fine to me either way. It's sort of a bummer to require synchronization now but it's not really the end of the world.

fitzgen: Removed the serialization at the cost of complicating `Type` with a boolean for whether or not it is for function entry blocks (which are implicit and therefore don't actually hold onto a `Type` in the serialization). I think this is the better choice though, and a bunch of stuff got simplified.

Issue #125, opened 2019-08-31
Fix errors reported by rust-clippy


lshlyapnikov:
### Describe the Bug

Running `cargo clippy` (https://github.com/rust-lang/rust-clippy) on `walrus` repository raises a bunch of warnings and two errors:
```
error: calls to `std::mem::drop` with a value that implements Copy. Dropping a copy leaves the original intact.
  --> src/module/elements.rs:80:21
   |
80 |                     drop(ty);
   |                     ^^^^^^^^
   |
   = note: #[deny(clippy::drop_copy)] on by default
note: argument has type wasmparser::primitives::Type
  --> src/module/elements.rs:80:26
   |
80 |                     drop(ty);
   |                          ^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#drop_copy

error: calls to `std::mem::drop` with a value that implements Copy. Dropping a copy leaves the original intact.
   --> src/module/elements.rs:227:13
    |
227 |             drop((id, segment));
    |             ^^^^^^^^^^^^^^^^^^^
    |
note: argument has type (id_arena::Id<module::elements::Element>, &module::elements::Element)
   --> src/module/elements.rs:227:18
    |
227 |             drop((id, segment));
    |                  ^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#drop_copy
```
The error message is self-explanatory.

### Steps to Reproduce

1. Install rust-clippy: https://github.com/rust-lang/rust-clippy
```
$ rustup component add clippy
```
2. Run rust-clippy on `walrus` codebase:
```
$ cd walrus/
$ cargo clippy
```
3. Search for ```error: calls to `std::mem::drop` with a value that implements Copy```

### Expected Behavior

rust-clippy does not report any errors.

### Actual Behavior

rust-clippy reports warnings and at least 2 errors.

### Additional Context

Would be nice to have no warning.

fitzgen:
Good catch -- would you like to send a pull request fixing this?

lshlyapnikov:
I will have look. Might have some questions.

PR #126, opened 2019-09-09
Fixing errors reported by `cargo clippy`

lshlyapnikov:
removing unnecessary `drop` calls that clippy reported as errors

See: #125


lshlyapnikov:
@fitzgen please have a look. Does not look like we need to do anything else unless we want to remove the corresponding entries from the underlying collection.

alexcrichton:
Looks great, thanks!

PR #133, opened 2019-10-10
Switch from `failure` to `anyhow`

alexcrichton:
This moves us back to `std::error::Error` compatibility and helps build
times since it doesn't rely on an older version of `syn`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ path = "benches/benches.rs"
 harness = false
 
 [dependencies]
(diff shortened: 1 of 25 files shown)


fitzgen:
https://travis-ci.org/rustwasm/walrus/jobs/596675888#L861-L865

```
error[E0599]: no method named `iter_chain` found for type `anyhow::error::Error` in the current scope
 --> examples/round-trip.rs:8:20
  |
8 |         for c in e.iter_chain() {
  |                    ^^^^^^^^^^
error: aborting due to previous error
```

Need to update the examples too.

PR #139, opened 2019-11-19
Allow custom sections to specify gc roots

alexcrichton:
This commit adds APIs necessary to allow custom sections to reference
core wasm functions/memories/etc and provide gc roots into them. This is
intended to be used, for example, with wasm interface types.

### DIFF
--- a/src/module/custom.rs
+++ b/src/module/custom.rs
@@ -1,5 +1,6 @@
 //! Working with custom sections.
 
+use crate::passes::Roots;
 use crate::tombstone_arena::{Id, Tombstone, TombstoneArena};
 use crate::CodeTransform;
 use crate::IdsToIndices;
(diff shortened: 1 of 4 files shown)


alexcrichton:
An example PR which consumes this is at https://github.com/bytecodealliance/wasm-interface-types/pull/14

src/module/custom.rs
```diff
@@ -29,6 +30,16 @@ pub trait CustomSection: WalrusAny + Debug + Send + Sync {
     /// payload. `walrus` will handle these for you.
     fn data(&self, ids_to_indices: &IdsToIndices) -> Cow<[u8]>;
 
+    /// Add any core wasm roots to the provided `roots` argument.
+    ///
+    /// This function will add any referenced core wasm items into the `Roots`
+    /// array provided.
+    ///
+    /// The default provided method does nothing.
+    fn gc_add_roots(&self, roots: &mut Roots) {
```
fitzgen: nitpick: let's name this `add_gc_roots`, since that makes more sense when read out loud in a sentence.

PR #141, opened 2019-11-26
Expose ImportedFunction

### LINKED ISSUE #140: ImportedFunction is not properly exposed
### Describe the Bug

`ImportedFunction` is used in the public API but its visibility not propagated correctly to the consumers.

### Steps to Reproduce

The easiest way to spot this is in docs:
image

As you can see, there are methods that return `ImportedFunction`, but it's not visible to a consumer in the same way as e.g. `LocalFunction` is.

RReverser:
Fixes #140.


alexcrichton:
Thanks!

PR #158, opened 2020-02-28
Add LICENSE files to match Cargo.toml

crjohns:
This change adds the LICENSE file text that matches Cargo.toml,
including a symlink in libraries under "crates".


alexcrichton:
Whoops, thanks!

PR #167, opened 2020-06-23
bump walrus-macro dep version

a1trl9:
Not sure if `walrus-macro` version in `dependencies` should be bumped to `0.17.0`.  I encountered the same error as in #165


alexcrichton:
Oops, thanks!

PR #185, opened 2020-10-27
Update wasmparser and pick up new SIMD instructions

alexcrichton:


alexcrichton:
cc @fitzgen, this removes the last vestiges of wasm validation from `walrus` since they were a bit out of alignment with the spec on function references anyway. All the validation happens in wasmparser now anyway, so these were a bit redundant and only implemented some validation, not everything with instructions.

fitzgen (approved):
Nice! :fire: down the custom validation!

PR #206, opened 2021-03-31
Bump wasmparser + support emitting multiple memories MemArgs

bnjbvr:
The first commit bumps wasmparser. I've made the minimal changes so that it compiles and runs, but most of the new SIMD / exception handling stuff is left to be implemented later.

The second commit allows emitting MemArgs that deals with memories that aren't the primary (0th) memory, by setting the right flag. I've confirmed it works as expected when playing with it in Wasmtime.

Closes #205.


alexcrichton:
Thanks for this!

PR #209, opened 2021-04-20
More permissive parsing of the "name" section

samsartor:
I've been working on getting DWARF debug symbols setup for a very large C++/emscripten app. Sentry.io's [symbolic](https://github.com/getsentry/symbolic) project is currently barfing on the bundle. Doing some investigation, I found that walrus was failing to parse the "name" custom section because emscripten appears to be:
1. leaving in symbols for deleted functions
2. implementing a bit of the https://github.com/WebAssembly/extended-name-section proposal

Since the "name" section is just debug info, I figured it would make sense to do more of a best-effort parsing, where we skip over any malformed entries instead of bailing out of the entire section on error. Looking at the `parse_name_section` function, there is already one such case (empty local names).

Originally I just added a bunch of match statements. But noticing there was a common pattern, I wrote a macro. If the macro is too scary, I can go back to the more verbose version.

### DIFF
--- a/src/module/mod.rs
+++ b/src/module/mod.rs
@@ -37,6 +37,7 @@ pub use crate::module::tables::{ModuleTables, Table, TableId};
 pub use crate::module::types::ModuleTypes;
 use crate::parse::IndicesToIds;
 use anyhow::{bail, Context};
+use log::warn;
 use std::fs;
 use std::mem;
 use std::path::Path;
@@ -371,9 +372,27 @@ impl Module {
         names: wasmparser::NameSectionReader,
         indices: &IndicesToIds,
     ) -> Result<()> {
+        // Like ? but `continue`s the current loop and prints a warning, rather
+        // than returning from the parse_name_section function.
+        //
+        // This is important because different compilers (notably emscripten)
+        // put a lot of garbage into this section, which we want to parse on a
+        // best-effort basis.
+        macro_rules! try_continue {
+            ($x:expr) => {
+                match $x {
+                    Ok(x) => x,
+                    Err(e) => {
+                        warn!("when parsing name section: {}", e);
+                        continue;
+                    }
+                }
+            };
+        }
+
         log::debug!("parse name section");
         for name in names {
-            match name? {
+            match try_continue!(name) {
                 wasmparser::Name::Module(m) => {
                     self.name = Some(m.get_name()?.to_string());
                 }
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for this! I think it makes sense to not give a full error when parsing the name section, but I think it's probably best to halt parsing the whole section after an error rather than trying to proceed as much as possible, could the error be caught by the caller and not propagated?

samsartor:
> Could the error be caught by the caller and not propagated?

It is actually! But because this code is bailing near the beginning of the name section, it winds up with a lot of data missing later on.

> I think it's probably best to halt parsing the whole section after an error rather than trying to proceed as much as possible

I am doing my best to bail on fatal errors. For example, see the `?`s remaining on any sort of "parse this integer" functions. The 3 places I'm replacing `?` with `try_continue!` are:
1. If the name section has an unknown subsection (e.g. https://github.com/WebAssembly/extended-name-section)
2. If the name section references a nonexistent function
3. If the name section references a nonexistent local

I'd certainly consider 2 and 3 non-fatal. It might be good to report them to the user somehow other than in logs, but they don't effect parsing at all. Unknown subsections should also be non-fatal since subsections include their length to enable parser recovery. However, looking again now, I admit I may not have implemented that correctly. What I'd really like is for wasmparser to add `NameType::Other` so that "I don't know what this subsection is" can be treated specially and separate from any other `BinaryReaderError`. Let me know if that change is needed to make this PR viable.

TBH, the `try_continue` macro is probably making this more confusing, not less. I'll see if plain-old match statements make it more clear that these locations don't really represent parser errors and are simply failures to deliver the encoded debug info to the desired object.

samsartor:
I confirmed that wasmparser doesn't correctly handle unknown subsection recovery in `NameSectionReader::read`. I'm not sure why that didn't turn up in my testing, but it means that I should remove 1 from this PR until I can get a wasmparser PR merged as well. I'd still like to see 2 and 3 get in.

alexcrichton:
Aha I have misinterpreted! I see now that `?` is still used in most places.

In that case, can this switch to using an explicit `match` instead of a macro? Additionally  could each match have a  comment indicating why it's ok to continue on failure?

Also if you're up for it then it'd be great to implement the extended name section here too :)

alexcrichton:
Looks great to me, thanks!

PR #231, opened 2022-06-29
transform DWARF code address

nokotan:
This is experiment implementation of transforming DWARF debug entries in WebAssembly binaries.

It seems to require more difficult implementation that using intermediate DWARF expression with [gimli.rs](https://github.com/gimli-rs/gimli), does not have direct access DWARF entries to be written.
This PR contains simpler way to transform code address:

- Instruction mapped address conversion
- Function code range based address conversion

Milestones

- [x] Function code range based address conversion
- [x] Implement instruction mapped address conversion
- [x] Better expression of code address that does not exactly match to instruction
  - beginning or end of function
- [x] Implement test cases

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ log = "0.4.8"
 rayon = { version = "1.1.0", optional = true }
 walrus-macro = { path = './crates/macro', version = '=0.19.0' }
 wasmparser = "0.78.0"
+gimli = "0.26.0"
 
 [features]
 parallel = ['rayon', 'id-arena/rayon']
--- a/crates/tests/Cargo.toml
+++ b/crates/tests/Cargo.toml
@@ -16,7 +16,7 @@ serde_json = { version = "1.0.40", features = ['preserve_order'] }
 tempfile = "3.1.0"
 walrus = { path = "../.." }
 walrus-tests-utils = { path = "../tests-utils" }
-wasmprinter = "0.2.25"
+wasmprinter = "=0.2.25"
 wat = "1.0.36"
 
 [features]
--- a/crates/tests/tests/custom_sections.rs
+++ b/crates/tests/tests/custom_sections.rs
@@ -86,8 +86,8 @@ fn smoke_test_code_transform() {
 
         fn apply_code_transform(&mut self, transform: &CodeTransform) {
             APPLIED_CODE_TRANSFORM.store(1, Ordering::SeqCst);
-            assert!(!transform.is_empty());
-            for (input_offset, output_offset) in transform.iter().cloned() {
+            assert!(!transform.instruction_map.is_empty());
+            for (input_offset, output_offset) in transform.instruction_map.iter().cloned() {
                 assert_eq!(input_offset.data() as usize + 3, output_offset);
             }
         }
--- a/src/emit.rs
+++ b/src/emit.rs
@@ -2,6 +2,8 @@
 //! structures. E.g. translating from globally unique identifiers down to the
 //! raw wasm structure's index spaces.
 
+use id_arena::Id;
+
 use crate::encode::{Encoder, MAX_U32_LENGTH};
 use crate::ir::Local;
 use crate::map::{IdHashMap, IdHashSet};
@@ -16,6 +18,7 @@ pub struct EmitContext<'a> {
     pub encoder: Encoder<'a>,
     pub locals: IdHashMap<Function, IdHashSet<Local>>,
     pub code_transform: CodeTransform,
+    pub function_ranges: Vec<(Id<Function>, wasmparser::Range)>,
 }
 
 pub struct SubContext<'a, 'cx> {
--- a/src/emit.rs
+++ b/src/emit.rs
@@ -2,8 +2,6 @@
 //! structures. E.g. translating from globally unique identifiers down to the
 //! raw wasm structure's index spaces.
 
-use id_arena::Id;
-
 use crate::encode::{Encoder, MAX_U32_LENGTH};
 use crate::ir::Local;
 use crate::map::{IdHashMap, IdHashSet};
@@ -18,7 +16,6 @@ pub struct EmitContext<'a> {
     pub encoder: Encoder<'a>,
     pub locals: IdHashMap<Function, IdHashSet<Local>>,
     pub code_transform: CodeTransform,
-    pub function_ranges: Vec<(Id<Function>, wasmparser::Range)>,
 }
 
 pub struct SubContext<'a, 'cx> {
--- a/src/ir/mod.rs
+++ b/src/ir/mod.rs
@@ -121,7 +121,7 @@ impl From<TypeId> for InstrSeqType {
 }
 
 /// A symbolic original wasm operator source location.
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Eq, Ord)]
 pub struct InstrLocId(u32);
 
 const DEFAULT_INSTR_LOC_ID: u32 = 0xffff_ffff;
--- a/src/module/debug.rs
+++ b/src/module/debug.rs
@@ -1,0 +1,383 @@
+use crate::emit::{Emit, EmitContext};
+use crate::{FunctionKind, ModuleCustomSections};
+use gimli::*;
+use std::borrow::Cow;
+use std::cmp::Ordering;
+use std::collections::BTreeMap;
+
+/// The set of de-duplicated types within a module.
+#[derive(Debug, Default)]
+pub struct ModuleDebugData {}
+
+impl ModuleDebugData {
+    /// c
+    pub fn convert_attributes<R: Reader<Offset = usize>>(
+        from_unit: read::Unit<R>,
+        dwarf: &read::Dwarf<R>,
+        convert_address: &dyn Fn(u64) -> Option<write::Address>,
+        convert_instrument_address: &dyn Fn(u64) -> Option<u64>,
+    ) -> Option<write::LineProgram> {
+        match from_unit.line_program {
+            Some(ref from_program) => {
+                let from_program = from_program.clone();
+                let line_program = ModuleDebugData::convert_line_program(
+                    from_program,
+                    dwarf,
+                    convert_address,
+                    convert_instrument_address,
+                )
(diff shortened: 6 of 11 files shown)


RReverser:
What's the status of this PR? From the description it sounds like it's almost ready, but OTOH it hasn't been changed from the draft state or updated since June...

I'm sure many people are eager to try this out once it's wired up all the way to wasm-bindgen.

nokotan:
Almost address conversions are implemented, especially DWARF debug line programs.
Some subprograms that removed away are still existent in DWARF entries, and high pc of a subprogram is incorrect because subprogram code size may be changed.

RReverser:
@alexcrichton @fitzgen Who is best positioned to review this?

alexcrichton:
Unfortunately I don't believe either of us can review this, and I'm not sure who else could. This crate is currently unmaintained at this time as well as neither of us is putting energy into it.

stan-irl:
> 



> > This crate is currently unmaintained at this time
> 
> Hmm considering how crucial it is to wasm-bindgen, that sounds mildly scary. Although I'm aware of the background and status of surrounding projects. That's unfortunate.
> 
> I suppose one of us could just try this PR with wasm-bindgen and, if it works, merge it anyway? Seems like @nokotan has done a great amount of work here and it's not like current state of DWARF works anyway, so it couldn't be worse.

This sounds like a reasonable path forward when there are no other options. The ecosystem needs this.

Liamolucko:
> I suppose one of us could just try this PR with wasm-bindgen and, if it works, merge it anyway? Seems like @nokotan has done a great amount of work here and it's not like current state of DWARF works anyway, so it couldn't be worse.

It seemed to work fine when I briefly tried it out.

RReverser:
> It seemed to work fine when I briefly tried it out.

Let's go ahead with it then?

71:
Thank you very much @nokotan for your work on this. I managed to make `wasm-bindgen` work with this pull request, and can now step through Rust code in the browser!

A few notes:
1. Since `wasm-bindgen` depends on `wit-walrus`, I had to fork `wit-walrus` to work with this PR: https://github.com/71/wasm-interface-types/commit/7d0392a2e584fc439063918c4ce96fa4ed981ed5.
   * `wit-*` crates are unmaintained since interface types are no longer being worked on. In theory `wasm-bindgen` could be modified to work without `wit-*`, **but** it seems like the WIT proposal is deeply integrated into the `wasm-bindgen` code, and undoing this would be a lot of work, especially since I only started doing this to debug another project of mine faster. I shall not yak-shave any longer.
2. I similarly had to fork `wasm-bindgen`, both to replace dependencies on `wit-walrus`, but also to replace dependencies on `walrus` directly: https://github.com/71/wasm-bindgen/commit/c2894739f1ffa20beb44e37d12bd9c56a2c442f0.
   * Simply depending on `git = "https://github.com/71/wasm-bindgen"` causes issues (e.g. with `console_error_panic_hook`), so I also need
     ```toml
     [patch.crates-io]
     js-sys = { git = "https://github.com/71/wasm-bindgen", package = "js-sys" }
     wasm-bindgen = { git = "https://github.com/71/wasm-bindgen" }
     ```
3. My browser (Chrome 110.0.5481.96) doesn't detect the embedded DWARF symbols (despite enabling them in "Experiments"), so I have to use [emscripten's `wasm-sourcemap.py`](https://github.com/emscripten-core/emscripten/blob/main/tools/wasm-sourcemap.py) to extract the DWARF symbols out of the `.wasm` output file after the `wasm-bindgen` pass. It also didn't work directly on my computer (problems mixing `bytes` and `str` in Python), so [I had to change it a little bit](https://gist.github.com/71/6346ad155d4df8544ce250d47880a1ee). The exact commands are:
   ```bash
   $ /path/to/forked/wasm-bindgen --out-dir ./wasm --out-name web --target web --weak-refs --keep-debug ./target/wasm32-unknown-unknown/debug/graph.wasm
   $ ./wasm-sourcemap.py ./wasm/web_bg.wasm --dwarfdump /path/to/bin/llvm-dwarfdump --output ./wasm/web_bg.wasm.map --basepath ./wasm -w ./wasm/web_bg.wasm --strip --source-map-url file:///absolute/path/to/wasm/web_bg.wasm.map
   ```
4. If I use `--reference-types` with `wasm-bindgen`, `llvm-dwarfdump` fails with `"error: web.wasm: invalid function export"`, so I cannot use `--reference-types`. `--weak-refs` work fine, though.
5. Symbol names aren't shown correctly, but this may be out of the scope of this PR. 
image

RReverser:
> Since `wasm-bindgen` depends on `wit-walrus`, I had to fork `wit-walrus` to work with this PR: [71/wasm-interface-types@7d0392a](https://github.com/71/wasm-interface-types/commit/7d0392a2e584fc439063918c4ce96fa4ed981ed5).

If you only want to replace dependencies, in theory it should be enough to use `[patch.crates-io]` section rather than fork each of those projects to get things up & running.

ortem:
Glad to see the work going on here, thank you for the great job @nokotan!

I've been playing with it recently too, and I'm glad to share that I've also managed to step through Rust code in Chrome browser with C/C++ DevTools Support extension enabled. It seems to work even with embedded DWARF symbols, so I didn't use `wasm-sourcemap.py`.

My steps:

1. Clone `wasm-bindgen` and `wasm-interface-types`, replace dependencies on `walrus` and `wit-walrus` (similar to @71)
2. Build a patched `wasm-bindgen`:
    ```sh
    $ cd wasm-bindgen
    $ cargo build --release --package wasm-bindgen-cli --bin wasm-bindgen
    ```
4. Build `examples/without-a-bundler` using `wasm-pack` with a patched `wasm-bindgen`:
    ```sh
    $ cd examples/without-a-bundler
    $ export PATH=/path/to/wasm-bindgen/target/release:$PATH
    $ wasm-pack build --debug --target web
    $ http
    ```

hello_from_rust

71:
Ah, indeed, it works without the need for source maps using the [C/C++ DevTools Support extension](https://chrome.google.com/webstore/detail/cc%20%20-devtools-support-dwa/pdcpmagijalfljmkmjngeonclgbbannb). It's even better than with source maps because local variables can be seen in the devtools as well. I'll note that it doesn't work on Chrome stable and that it requires a Chrome extension, so the source maps solution, though not as good, is more easily accessible.

@RReverser Ah right I didn't try to use `[patch.crates-io]` (initially I modified some things in my fork, but in the end I discarded these changes and only changed the imports). I'm not sure how that interacts with dependent packages, but this should be fine and [it does simplify things quite a bit](https://github.com/71/wasm-bindgen/commit/f6b20b55688c7a99c14e54bf516cbae24d81191e). Unfortunately we can't do that straight from the dependent package because `wasm-bindgen-cli` is installed globally, and unrelated to a project's dependencies.

ltfschoen:
i got it to work too, thanks everyone, particularly the guides here by @71 and @ortem and the guide by @michael8090 [here](https://github.com/rustwasm/wasm-bindgen/issues/2389#issuecomment-1071999106). i summarised my whole journey getting it to work [here](https://github.com/maciejhirsz/kobold/issues/47).

but i'm confused why i had to do this step https://github.com/maciejhirsz/kobold/issues/47#issuecomment-1493375173 to fix missing files. i think that may be because i copied the code that @71 did into my dependencies and maybe that uses source maps, and may be i didn't configure them.
i did that because when i was stepping-through the rust code i'd get errors that some files were not found, and it was because it was trying to access files in file://  /cargo and /rustc root path locations that didn't even exist and i am on macOS. [see this screenshot](https://user-images.githubusercontent.com/6226175/229325620-c20b8bdc-30f8-4b5e-82af-631bffa5a489.png).  
otherwise i had no issues stepping-through the files of project i was debugging in  file://  /Users/luke/.cargo and /Users/luke/code/github/ltfschoen/kobold, as shown

AlphaPrime8:
I (finally) got this to work by manually adding symbolic links on my local filesystem since the dwarf "line table" generated by default apparently had invalid paths. 

Note: I did not use the `wasm-sourcemap.py` script.

1. I simply added the patches to Cargo.toml referenced in: https://github.com/rustwasm/walrus/pull/231#issuecomment-1431438647
2. I build with `wasm-pack build --dev --target web --out-dir pkg` (i'm using [egui](https://github.com/emilk/egui) from a basic react app [cra5])
3. Note also I did not use --weak-refs, that made it break.
4. In picture below you can see how the default generated source map was trying to get source files, so I just sym-linked such that those paths became valid.

Screenshot from 2023-05-30 01-27-06

Also added this metadata in my Cargo.toml:
```
[package.metadata.wasm-pack.profile.dev]
wasm-opt = false

[package.metadata.wasm-pack.profile.dev.wasm-bindgen]
debug-js-glue = true
demangle-name-section = true
dwarf-debug-info = true
```

Next challenge will be to pass solana transactions from Rust to JS so they can be consumed by the wallet adapter.

guybedford:
I'm currently assisting with some basic maintenance on this project, and we have already merged some initial wasmparser upgrades and a migration to wasm-encoder.

I'd like to investigate this PR further towards a merge for the next release. I'll open a new rebase PR for this soon unless we can pick up on the rebase here.

stan-irl:
I still wasnt able to step through.  I hit a breakpoint and then after that the debugger just jumps to random places in random source files. 

I had the following config:
```
[patch.crates-io]
js-sys = { git = "https://github.com/71/wasm-bindgen", package = "js-sys" }
wasm-bindgen = { git = "https://github.com/71/wasm-bindgen" }

[package.metadata.wasm-pack.profile.dev]
wasm-opt = false

[package.metadata.wasm-pack.profile.dev.wasm-bindgen]
debug-js-glue = true
demangle-name-section = true
dwarf-debug-info = true
```

I also set a simple mapping from my rust toolchain to the expected path in the extension settings:
```
/rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library:~/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library
```

which seems to fix the missing sources

guybedford:
This landed in https://github.com/rustwasm/walrus/pull/244, available in 0.20.0. Any issues please do report back.

artemmukhin:
Glad to see it's finally landed. Kudos to @nokotan and @guybedford!

PR #232, opened 2022-07-15
fix test on Windows and fix CI failure

michael8090:
It fixes the path escape issue on Windows.

And by using a fixed version number for wasmprinter, it fixes the CI failure that has been there for some days.


guybedford:
This should be working now on main.

Issue #234, opened 2022-08-03
Floating point detection


reuvenpo:
### Summary

Does Walrus have an way to check if a wasm module contains any floating point operations?

### Additional Details

I'm looking for a way to (quickly?) detect if a wasm module is using floating points in any way. I don't see a method for it, but is there a way to easily check for this? Ideally the function would just return a boolean, I don't need a list of all floating point instructions.

PR #235, opened 2022-08-07
added public methods to iterate all blocks of a local function

reuvenpo:
This PR adds two methods to iterate over all the blocks in a LocalFunction. This allows making per-block transformations, or simple full-function analysis, without having to manually navigate the function's execution branches to find all reachable blocks. CC #234


reuvenpo:
seems like the CI is failing due to issues that are unrelated to my additions

PR #236, opened 2022-10-19
Fix failing round-trip tests

adambratschikaye:
Many of the round-trip tests seem to be broken on master (mainly due to formatting changes). This change fixes the expected output so all tests now pass.

These changes were mostly done automatically, first by changing tests that use the `CHECK` and `NEXT` directives to use `CHECK-ALL`:

```
for f in $(rg -l NEXT crates/tests/tests/round_trip/)
do
    sed -i 's/;; CHECK: /(; CHECK-ALL:\n/' $f;
    sed -i 's/;; NEXT://' $f;
    echo ';)' >> $f;
done
```

and then updating the out put of those tests with
```
WALRUS_BLESS=1 cargo test --all
```

### DIFF
--- a/crates/tests/tests/round_trip/anyref1.wat
+++ b/crates/tests/tests/round_trip/anyref1.wat
@@ -10,6 +10,8 @@
     (import "x" "y" (func (;0;) (type 0)))
     (func (;1;) (type 0) (param externref)
       local.get 0
-      call 0)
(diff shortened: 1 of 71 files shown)


guybedford:
Thanks for the PR, this has been landed in https://github.com/rustwasm/walrus/pull/243 thanks to @plusvic.

PR #242, opened 2023-05-29
refactor: use wasm-encoder dependency for all encoding

guybedford:
Upgrades all encoding to use the latest version of wasm-encoder 0.29.0.

This is only a refactoring, with no outward API changes.


src/module/mod.rs
```diff
@@ -352,13 +351,22 @@ impl Module {
...
+        let out = cx.wasm_module.finish();
         log::debug!("emission finished");
-        wasm
+
+        // let mut validator = Validator::new();
+        // if let Err(err) = validator.validate_all(&out) {
+        //     eprintln!("{:?}", err);
+        //     panic!("Unable to validate serialized output");
+        // }
+
```
elliottt, line 368: Any reason to keep this commented-out code?
```suggestion
```

elliottt (approved):
This all looks good to me!

src/module/mod.rs
guybedford: It's quite useful for testing, but it could be removed.

PR #245, opened 2023-06-16
Use backwards-compatible encoding of element segments where possible

Liamolucko:
This is to fix https://github.com/rustwasm/wasm-bindgen/actions/runs/5278136842/jobs/9547894589, which is happening because `@webassemblyjs/wasm-parser` (used by Webpack) doesn't support the newer encoding of element segments that `wasm-encoder` uses by default.

This was the behaviour prior to #242.

The alternative to this would be to update `@webassemblyjs/wasm-parser` to support the new encoding, but I think we may as well be backwards-compatible where possible anyway.

### DIFF
--- a/src/module/elements.rs
+++ b/src/module/elements.rs
@@ -186,23 +186,7 @@ impl Emit for ModuleElements {
                     })
                     .collect();
                 let els = wasm_encoder::Elements::Expressions(els_vec.as_slice());
-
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for the fix. I did run the wasm-bindgen tests against this release before making it, but seems I somehow missed this case. Posting a patch now.

PR #248, opened 2023-10-10
feat: upstream wasi virt walrus ops

vados-cosmonic:
This PR upstreams a few functions from WASI-Virt (see [previous discussion](https://github.com/bytecodealliance/WASI-Virt/issues/20)) that might be helpful to the wider `walrus` user ecosystem.

### DIFF
--- a/src/module/exports.rs
+++ b/src/module/exports.rs
@@ -1,5 +1,7 @@
 //! Exported items in a wasm module.
 
+use anyhow::Context;
+
 use crate::emit::{Emit, EmitContext};
 use crate::parse::IndicesToIds;
 use crate::tombstone_arena::{Id, Tombstone, TombstoneArena};
@@ -98,6 +100,16 @@ impl ModuleExports {
             ExportItem::Function(f0) => f0 == f,
             _ => false,
         })
+    }
+
+    /// Retrieve an exported function by name
+    pub fn get_func_by_name(&self, name: impl AsRef<str>) -> Result<FunctionId> {
+        self.iter()
+            .find_map(|expt| match expt.item {
+                ExportItem::Function(fid) if expt.name == name.as_ref() => Some(fid),
+                _ => None,
+            })
+            .with_context(|| format!("unable to find function export '{}'", name.as_ref()))
     }
 
     /// Get a reference to a table export given its table id.
@@ -352,6 +364,16 @@ mod tests {
         module.exports.delete(export_id);
 
         assert!(module.exports.get_exported_func(fn_id).is_none());
+    }
+
+    #[test]
+    fn get_func_by_name() {
+        let mut module = Module::default();
+        let fn_id: FunctionId = always_the_same_id();
+        let export_id: ExportId = module.exports.add("dummy", fn_id);
+        assert!(module.exports.get_func_by_name("dummy").is_ok());
+        module.exports.delete(export_id);
+        assert!(module.exports.get_func_by_name("dummy").is_err());
     }
 
     #[test]
--- a/src/module/functions/mod.rs
+++ b/src/module/functions/mod.rs
@@ -1,5 +1,15 @@
 //! Functions within a wasm module.
 
+use std::cmp;
+use std::collections::BTreeMap;
+
+use anyhow::{bail, Context};
+use wasm_encoder::Encode;
+use wasmparser::{FuncValidator, FunctionBody, Range, ValidatorResources};
+
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
+
 mod local_function;
 
 use crate::emit::{Emit, EmitContext};
@@ -11,18 +21,18 @@ use crate::parse::IndicesToIds;
 use crate::tombstone_arena::{Id, Tombstone, TombstoneArena};
 use crate::ty::TypeId;
 use crate::ty::ValType;
-use std::cmp;
-use std::collections::BTreeMap;
-use wasm_encoder::Encode;
-use wasmparser::{FuncValidator, FunctionBody, Range, ValidatorResources};
-
-#[cfg(feature = "parallel")]
-use rayon::prelude::*;
+use crate::{ExportItem, Memory, MemoryId};
 
 pub use self::local_function::LocalFunction;
 
 /// A function identifier.
 pub type FunctionId = Id<Function>;
+
+/// Parameter(s) to a function
+pub type FuncParams = Vec<ValType>;
+
+/// Result(s) of a given function
+pub type FuncResults = Vec<ValType>;
 
 /// A wasm function.
 ///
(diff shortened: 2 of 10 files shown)


src/module/functions/mod.rs
```diff
@@ -418,6 +433,66 @@ impl Module {
...
+    ///
+    /// When called, if `builder` produces a None value, the function in question will be
+    /// replaced with a stub that does nothing (more precisely, a function with an unreachable body).
+    pub fn replace_fn_by_name<F>(&mut self, name: impl AsRef<str>, builder: F) -> Result<()>
+    where
+        F: FnOnce() -> Result<Option<Function>>,
+    {
+        let built_fn = builder().context("fn builder failed")?;
+
+        // TODO: Hmnn.... how should we distinguish between which one to go for?
```
guybedford, line 489: I'd suggest having this function take a function id instead to resolve this.

src/module/functions/mod.rs
```diff
@@ -206,6 +210,17 @@ impl ModuleFunctions {
         self.arena.delete(id);
     }
 
+    /// Remove a function from this module by name that corresponds to it's import or export.
+    ///
+    /// Unlike [`by_name()`](ModuleFunctions::by_name), "name" corresponds to either a functions import name or it's export name.
+    pub fn delete_by_name(&mut self, name: impl AsRef<str>) {
```
guybedford, line 216: Again, perhaps this should be by Id.

guybedford (commented):
Looks like a great start, these seem like useful conveniences to me.

src/module/functions/mod.rs
```diff
@@ -418,6 +433,66 @@ impl Module {
...
+            })
+            .nth(0)
+            .with_context(|| format!("unable to find function export '{}'", name.as_ref()))
+    }
+
+    /// Retrieve the ID for the first exported memory.
+    ///
+    /// This method does not work in contexts with [multi-memory enabled](https://github.com/WebAssembly/multi-memory),
+    /// and will error if more than one memory is present.
+    pub fn get_primary_memory_id(&self) -> Result<MemoryId> {
```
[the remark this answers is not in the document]
vados-cosmonic: Will fix! "primary" shouldn't have been in there, I realized that wasn't canonical half way through. 

- [ ] rename `get_primary_memory_id` to `get_memory_id`

src/module/functions/mod.rs
vados-cosmonic: Will do, that's a good way to make it explicit.

- [ ] Take ID for replacing function

src/module/functions/mod.rs
guybedford: Right then perhaps delete is superfluous then. Having them on `ModuleFunctions` seems preferable to me. And that also allows them to have shorter names.

vados-cosmonic: Thanks for the feedback, will get to integrating this. I filled out more of the other implementations and I need to do some looking and thinking, should have something ready for review tomorrow!

- [ ] Move all func-y modules to `ModuleFunctions`

guybedford (commented):
Thanks @vados-cosmonic, this is mostly seeming pretty good to me.

src/module/exports.rs
```diff
@@ -100,6 +102,16 @@ impl ModuleExports {
         })
     }
 
+    /// Retrieve an exported function by name
+    pub fn get_func_by_name(&self, name: impl AsRef<str>) -> Result<&Export> {
```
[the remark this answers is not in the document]
vados-cosmonic: Yeah I wanted to match the other functions here that seem to deal only in `&Export`/`Option<&Export>` but I will change this to return the `FunctionId`

- [ ] change `get_func_name` in exports to return `FunctionId`

src/module/imports.rs
```diff
@@ -103,6 +104,35 @@ impl ModuleImports {
...
+    /// Retrieve an imported function by ID
+    pub fn get_func_by_id(&self, id: FunctionId) -> Option<&Import> {
+        self.arena.iter().find_map(|(_, import)| match import.kind {
+            ImportKind::Function(fid) if fid == id => Some(import),
+            _ => None,
+        })
+    }
+
+    /// Check whether imports include a function with the given ID
+    pub fn contains_function_by_id(&self, fid: FunctionId) -> bool {
```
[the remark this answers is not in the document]
vados-cosmonic: Done -- I can definitely remove this!

- [ ] remove `contains_function_by_id`

src/module/memories.rs
```diff
@@ -115,6 +115,11 @@ impl ModuleMemories {
     pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut Memory> {
         self.arena.iter_mut().map(|(_, f)| f)
     }
+
+    /// Get the number of memories in this module
+    pub fn len(&self) -> usize {
```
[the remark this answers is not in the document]
vados-cosmonic: :+1: will leave this in

src/module/functions/mod.rs
vados-cosmonic: Hey @guybedford what did you think about the `replace_fn_by_id` stuff here? That's the biggest hairiest fn, and it just seems to be a little bit messier than the others.  

If you've got any ideas for refactoring, I'd love to hear them -- it's a little bit messy as-is, I'm not a huge fan of just *trying* to get the function from the imports, as opposed to having two instances of `replace_func_by_id` (one for `ModuleImports` another for `ModuleExports`). 

The unified implementation just ends up being basically two big `if`s that do just-slightly-different things for imports/exports because of how the types/destructuring works out.

`module.exports.replace_func_by_id(...)` and `module.imports.replace_func_by_id(...)` doesn't seem so bad versusa `module.replace_func_by_id(...)` which is implemented as it is now.

vados-cosmonic:
Thanks for all the work reviewing this, improving the function/usage surface here has certainly taken more work than I expected.

src/module/functions/mod.rs
guybedford: Yeah I appreciate it's got two completely separate branches here. What happens if the function is not an import or an export, does it work correctly then as well?

vados-cosmonic: At present, it throws an error like this:

```rust
bail!("function with ID [{fid:?}] is neither an import or an export");
```

Is that right? I was assuming that we want to fail if a function is neither an import nor export.

Assuming it's "right" It's not terrible, but it's also not *great* either, I think.

The big thing that worried me which was missing was the "upsert" semantic of [the similar downstream function in WASI-virt](https://github.com/bytecodealliance/WASI-Virt/blob/main/src/walrus_ops.rs#L130). Since the original functions were by name, it had the ability to *insert* the function if it was missing.

This means I couldn't perfectly replace that stubbing functionality (but of course, someone could write quite succinct code that checked for the existing function and replaced as necessary, something like:

```rust
if let Some(fid) = module.exports.get_func_by_name(export_name) {
    replace_func_by_id(...);
}  else {
    module.exports.add(export_name, ...) // function builder stuff has to get duplicated here, unfortunately
}
```

So to summarize, it *feels* like:
- the functions should be manipulated by name (so we can preserve that upsert functionality)
- the replacement function should probably live on the individual imports/exports objects

guybedford: Yeah I think this function as framed as a kind of replacer by id, only makes sense if it also works for arbitrary locals? But then perhaps I've framed this wrong and focusing on the separate import and export cases is better?

It would be fine to just wrap a get_or_create_export functionality around it if we could figure out a way to make it work properly though I think for the WASI-Virt case.

I'm open to suggestions though.

vados-cosmonic: > It would be fine to just wrap a get_or_create_export functionality around it if we could figure out a way to make it work properly though I think for the WASI-Virt case.

I do like adding a `get_or_create_export` so WASI-Virt can be cleaner... What about a mash of the solutions?

`ModuleExports::replace_func_by_id` (or `replace_func()` ?
`Module::get_or_create_export`

`ModuleImports::replace_func_by_id`
`Module::get_or_create_import`

There's a bit of duplication, but I think it's definitely least-surprise.

vados-cosmonic: oh also, I haven't forgotten about tests! Will add after we've figured out a shape that feels reasonable

guybedford: Okay, I'm open to that if you would prefer. I think for this kind of a `replace_func_by_id` to work it would absolutely need to handle functions that aren't exported or imported.

vados-cosmonic: Yeah it's not ideal but the reason I'm thinking `replace_func_by_id` (if it worked for local functions) would be troublesome is that if you tried to use it from `get_or_create_import` it *could* be misused and replace a local function by accident.

I'll think about it a bit more -- please let me know if you have any ideas on how to write it in a way that could work for either... 

I mean it's overkill but we could have those function *AND* a `Module::replace_func_by_id` which uses the other two (and handles local functions). Still seems possible to misuse, but we can have the `module.imports.replace_func...` and `module.replace_func...` which works for everything.

guybedford: Yeah, perhaps let's just focus on the separate import and export cases for now.

vados-cosmonic:
Hey @guybedford I've done a little bit more on this -- split up the functions and added some basic tests, if you wouldn't mind taking another look

src/module/functions/mod.rs
```diff
@@ -418,6 +428,124 @@ impl Module {
...
+                // Run the builder function to produce a new local function, or create a stub that is unreachable
+                let new_local_fn = match fn_builder((&params, &results))
+                    .context("import fn builder failed")?
+                {
+                    Some(f) => f,
+                    None => {
+                        let mut builder = FunctionBuilder::new(&mut self.types, &params, &results);
+                        builder.func_body().unreachable();
+                        builder.name("no-op".into());
+                        builder.local_func(vec![])
```
guybedford, line 534: I'd prefer we keep stubbing as an implementation detail rather than providing this behaviour by default. We can definitely add an example of stubbing though in the comments if you like.

src/module/functions/mod.rs
```diff
@@ -418,6 +428,124 @@ impl Module {
...
+
+                // Run the builder function to produce a new local function, or create a stub that is unreachable
+                let new_local_fn = match fn_builder((&params, &results))
+                    .context("export fn builder failed")?
+                {
+                    Some(f) => f,
+                    None => {
+                        let mut builder = FunctionBuilder::new(&mut self.types, &params, &results);
+                        builder.func_body().unreachable();
+                        builder.local_func(vec![])
```
guybedford, line 482: Again, I think we should keep stubbing to being an implementation detail.

src/module/functions/mod.rs
```diff
@@ -418,6 +428,124 @@ impl Module {
...
+            .context("module does not export a memory")
+    }
+
+    /// Replace a single exported function with the result of the provided builder function.
+    ///
+    /// When called, if `builder` produces a None value, the function in question will be
+    /// replaced with a stub that does nothing (more precisely, a function with an unreachable body).
+    ///
+    /// This function returns the function ID of the *new* function
+    pub fn replace_exported_func_by_id<F>(
```
guybedford, line 454: Perhaps just `replace_exported_func` instead of `by_id` for a little brevity? (and for imports)

guybedford (approved):
Looks great, thank you. Just a note re naming and I think we should keep stubbing to being a user side implementation detail.

src/module/functions/mod.rs
vados-cosmonic: That's reasonable -- will update this and add an example to the documentation (maybe even a doc test?)

- [ ] remove stubbing as default behavior

src/module/functions/mod.rs
vados-cosmonic: Thanks, will change it here as well

- [ ] remove stubbing

src/module/functions/mod.rs
vados-cosmonic: Yup, I'm definitely good with that -- will update

- [ ] rename `replace_exported_func_by_id` to `replace_exported_func`

vados-cosmonic:
Addressed all comments and rebased! Thanks for the reviews -- I think you should be good to merge at will

guybedford:
@vados-cosmonic this looks great to merge to me as well. I'm not sure what is up with the recent CI break, given nothing has changed on that side. If you have any ideas do let me know, I think we should ensure CI is running before landing.

vados-cosmonic:
Hey @guybedford I think I found the problem -- the WAT used in the tests *seem* to be outdated in two ways:

- `get_local` (that's now `local.get`, right?)
- s-expressions in WAT being invalid (maybe the syntax without making the body a proper s-expressions is supported then/now, but the easy fix was to add them)

For reference, the errors in CI look like this:

```
running 5 tests
test tests_invalid_multi_value_0_wat ... FAILED
test tests_invalid_multi_value_1_wat ... FAILED
test tests_invalid_if_with_no_else_wat ... ok
test tests_invalid_multi_value_2_wat ... ok
test tests_invalid_multi_value_3_wat ... ok

failures:

---- tests_invalid_multi_value_0_wat stdout ----
thread 'tests_invalid_multi_value_0_wat' panicked at crates/tests-utils/src/lib.rs:107:23:
got an error: unknown operator or unexpected token
     --> tests/invalid/multi-value-0.wat:5:5
      |
    5 |     get_local 0))
      |     ^
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests_invalid_multi_value_1_wat stdout ----
thread 'tests_invalid_multi_value_1_wat' panicked at crates/tests-utils/src/lib.rs:107:23:
got an error: unknown operator or unexpected token
     --> tests/invalid/multi-value-1.wat:4:5
      |
    4 |     get_local 0
      |     ^


failures:
    tests_invalid_multi_value_0_wat
    tests_invalid_multi_value_1_wat
```

It seems to be simply a WAT format/issue... I'm not sure how this could have possibly been in here so long without causing tests to fail..

The failing tests were written in 2019, so I'm hoping it's *actually* [this simple of a fix](https://github.com/rustwasm/walrus/pull/248/commits/45cbc48fc3b632bf8e9c3f20c1b4bfba4c276eac). 

I'm going around and updating the WAT as best I can, will collect them into one separate commit so they're easy to review!

guybedford:
I'm going to merge this, but before we release I would still like to see how the Wasi-VIRT rebase looks like on top of this work to see if there's any final API feedback we need to make in follow-up PRs.

Thanks for getting this over the line!

vados-cosmonic:
Hey @guybedford sure -- I'll have a PR up to WASI-Virt and try and use these new functions so we have an idea of how different it will look!

PR #249, opened 2023-10-14
fix(tests): update WAT used in tests

vados-cosmonic:
The excellent pre-existing test suite contains a lot of WAT files as test-cases that seem to have older syntax in them.

The breakage is introduced because the current test suite runs stable, latest and nightly versions of Rust while conducting tests. While the rest of the Wasm toolchain (binaryen, wabt, etc) are hard-coded to versions that should stay compatible with the WAT as they were written, the upstreaming of various changes to Rust itself seems to be causing failures when using the existing WAT for tests.

This PR updates the WAT (mostly operation renaming) that have changed in order to get the [`stable`|`beta`|`nightly`] tests passing again.


guybedford:
Thank you so much @vados-cosmonic, great to see CI green again.

PR #250, opened 2023-10-14
refactor: improve ergonomics of fn replace

vados-cosmonic:
After trialing `replace_(imported|exported)_func` in WASI-Virt, it's clear that the ergonomics around the builder function need to be improved. `FunctionBuilder` (particularly `FunctionBuilder::new()`) is difficult to use without a mutable borrow of the module itself.

This commit refactors `replace_(imported|exported)_func` in order to pass through the mutable borrow which makes it easier to use `FunctionBuilder`s.

### DIFF
--- a/src/module/functions/mod.rs
+++ b/src/module/functions/mod.rs
@@ -21,7 +21,7 @@ use crate::parse::IndicesToIds;
 use crate::tombstone_arena::{Id, Tombstone, TombstoneArena};
 use crate::ty::TypeId;
 use crate::ty::ValType;
-use crate::{ExportItem, Memory, MemoryId};
+use crate::{ExportItem, FunctionBuilder, InstrSeqBuilder, LocalId, Memory, MemoryId};
 
 pub use self::local_function::LocalFunction;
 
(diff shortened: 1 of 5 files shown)


guybedford:
Thanks for putting some more thought to this. I was wondering - could we just pass an already-created `InstrSeqBuilder` here so that the first argument to the function is `body` where `body` is the builder for the function builder? Just like a `|then|` works today - https://docs.rs/walrus/0.20.1/walrus/struct.InstrSeqBuilder.html#example-4.

vados-cosmonic:
Thanks for the pointer, that's a great idea -- I'll make the change!

- [ ] Use `impl IntrSeqBuilder` modification fn similar to `if_else` on `FunctionBuilder`

vados-cosmonic:
Hey @guybedford Ran into an issue -- with [the changes](https://github.com/rustwasm/walrus/pull/250/commits/359fb1c70a8d64c04dd8f8f29f53b64a70ebfbe0), I run into an issue where I can't get the `LocalFunction` to be returned from the closure. 

The way the other functions work, they don't generate the `LocalFunction`, they just let you manipulate the `InstrSeqBuilder`, but in this case we *want* the user to be able to pass the arguments right? So that means they need to actually build the local function.

Here's a snapshot of the problem:

```rust
        // Replace the existing function with a new one with a reversed const value
        let new_fn_id = module
            .replace_exported_func(original_fn_id, |mut body| {
                body.i32_const(4321).drop();
                body.local_func(vec![]) // TODO(FIX): Can't move LocalFunction produced here out of deref to body!
            })
            .expect("function replacement worked");
```

We want the user to have control over that `local_func` fall to actually generate the function with the arguments they want (ex. maybe you want to always give the same arguments to some fn), but we get this error:

```rust
error[E0507]: cannot move out of dereference of `InstrSeqBuilder<'_>`
   --> src/module/functions/mod.rs:706:17
    |
706 |                 body.local_func(vec![]) // TODO(FIX): Can't move LocalFunction produced here out of deref to body!
    |                 ^^^^^^^^^^^^^^^^^^^^^^^ move occurs because value has type `FunctionBuilder`, which does not implement the `Copy` trait
```
This also is an issue if we pass in the full `FunctionBuilder` (rather than the seq builder), rather than create it inside.

guybedford:
> in this case we want the user to be able to pass the arguments right?

Could we just use the same localIds from the original function, passing that list as one of the arguments?

vados-cosmonic:
Ah in that case, we could call `body.local_func(..)` from outside all-together, that would work, I think. I'll give it a shot

guybedford:
Unless there are other drawbacks we need to be aware of it seems like that might be the idiomatic pattern to me? The only case I think where it might be restrictive is if you want to pass an already constructed function?

vados-cosmonic:
Yeah I think that's right -- IMO being able to pass an already pre-created function is pretty useful, but that's a bridge that we can cross when we get there. I don't know exactly how but , but there's probably a clean way to graft the implementation of an existing `Function` to another one using `append_instruction` and the `instruction_mapping` in there, so you could graft all the instructions over fairly easily in the closure.

Going to take one final look and this should be ready for review!

src/module/functions/mod.rs
```diff
@@ -815,14 +817,10 @@ mod tests {
...
-        let mut builder = FunctionBuilder::new(&mut module.types, &[], &[]);
-
         // Replace the existing function with a new one with a reversed const value
         let new_fn_id = module
-            .replace_imported_func(original_fn_id, |_| {
-                builder.func_body().unreachable();
-                Ok(builder.local_func(vec![]))
+            .replace_imported_func(original_fn_id, |body| {
+                body.unreachable();
             })
```
guybedford, line 824: Remind me what the pattern is to determine the args local ids here? Can we pass them as secondary arguments like `|(body, args)|` kind of thing? Or is there another way?

vados-cosmonic: Ah, so I ripped the argument stuff out -- the reason it was in there was so that people who created `FunctionBuilder`s would have the right params/results (to use), but since that now happens [before the builder fn runs](https://github.com/rustwasm/walrus/pull/250/files#diff-3848d3510e4aff5047c97fce1309a78bcdcef84dcc5d5e887d61d51bb8d19af1R487), I don't even have to pass that in anymore.

we *can* pass a pointer to the params & results if we want, though, if you think it is helpful to people who are writing the bodies.

vados-cosmonic:
Sorry a small formatting issue slipped through, squashed & it's ready when tests pass & you're good to merge :bow:

src/module/functions/mod.rs
guybedford: Yeah I get that the args and result types aren't available, but what I mean is the `LocalId` references for the args, not the actual information about the args. It might even be nice to have all the info though I think too `|(body, arg_locals, args, results)|`.

Good API design is hard :) But just making sure we don't miss the important `LocalId` info to be able to read an arg without requiring another lookup of the function.

guybedford: My remaining question is just this: if replacing a function that takes an argument, the function localids will remain the same. How do I lookup those function local ids? Do I just assume that the first n locals correspond to the n args? Does that invariant always hold?

Just want to ensure we aren't leaving a gap. Having an example of reading a function argument in the tests would also help.

src/module/functions/mod.rs
```diff
@@ -447,98 +447,112 @@ impl Module {
...
+        {
+            // Retrieve the params & result types for the exported (local) function
+            let ty = self.types.get(lf.ty());
+            let (params, results) = (ty.params().to_vec(), ty.results().to_vec());
+
+            // Add the function produced by `fn_builder` as a local function
+            let mut builder = FunctionBuilder::new(&mut self.types, &params, &results);
+            let mut new_fn_body = builder.func_body();
+            builder_fn(&mut new_fn_body);
+            let func = builder.local_func(lf.args.clone());
```
guybedford, line 490: `lf.args()` is the local list I'm referring to fwiw.

src/module/functions/mod.rs
```diff
@@ -447,98 +447,112 @@ impl Module {
...
+        {
+            // Retrieve the params & result types for the imported function
+            let ty = self.types.get(*tid);
+            let (params, results) = (ty.params().to_vec(), ty.results().to_vec());
+
+            // Build the new function
+            let mut builder = FunctionBuilder::new(&mut self.types, &params, &results);
+            let mut new_fn_body = builder.func_body();
+            builder_fn(&mut new_fn_body);
+            let new_func_kind = FunctionKind::Local(builder.local_func(vec![]));
```
guybedford, line 544: Are no local Ids being passed here? We should pre-generate a new `localId` for each argument here I think?

guybedford (changes requested):
More clearly marking the request for changes per the query in comment https://github.com/rustwasm/walrus/pull/250#discussion_r1364250475.

src/module/functions/mod.rs
vados-cosmonic: Ahhh I see what you mean, thank you for explaining -- I agree with what having that information would be helpful here... I think it *would* be hard to get it with only the body available.

I'm not sure exactly how you *would* get those function local IDs actually, you'd have to basically remember what the parameters were...

vados-cosmonic: Ah I can work with that -- yeah I'll make this change to pass through a pointer to the arg locals. Do you think it wouldn't be useful to pass through the params/results? I know that they're generally already available in some sense but just want to make sure on that

src/module/functions/mod.rs
vados-cosmonic: Yeah this was a bit confusing to me, but I looked back at the previous implementations and saw that they were actually always empty.... So I thought that *might* work here. 

One problem I had here was that there *is* no lcoal func to get args from here right? Am I misunderstanding maybe?

src/module/functions/mod.rs
vados-cosmonic: Ah thanks this is helpful -- I'll pass it through from here

src/module/functions/mod.rs
guybedford: Yeah in this line you exactly need to be creating the localIds - because otherwise it won't be possible to read arguments for the created function.

vados-cosmonic: Ahhh gotcha, thank you -- I didn't understand what you meant by generating the new IDs *per argument*. In the other case we can just use the args but here we have to *create* them, rather than the person being able to (though they theoretically could themselves if they knew which function to replace). 

Will update this bit too

vados-cosmonic: OK so I *think* I have it right here -- https://github.com/rustwasm/walrus/pull/250/commits/af96691e6b286dd8847474aee5e732764eff7bdd -- I'm generating new `LocalId`s to match each param w/ the right internal type.

guybedford (approved):
Yeah this looks right to me now! Let's land once the tests are passing.

vados-cosmonic:
Ahhh apologies I forgot to convert some more spots to the new API. tests have been fixed

PR #252, opened 2023-10-23
rename: get_func_by_name to get_func

guybedford:
This renames the new `(imports|exports).get_func_by_name` to just `(imports|exports).get_func`. The main reason being that I think the `by_name` term is overloaded in the functions list already to mean the actual function name instead of the import or export name, so wanted to avoid any confusion here.

Then I updated the `delete_by_name` functions to be generic named `remove` functions also without the `by_name` term to avoid confusion about it.

### DIFF
--- a/src/module/exports.rs
+++ b/src/module/exports.rs
@@ -102,8 +102,8 @@ impl ModuleExports {
         })
     }
 
-    /// Retrieve an exported function by name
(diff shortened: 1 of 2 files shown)


vados-cosmonic:
Looks good to me -- thanks for the pointer, updating the WASI-Virt PR now!

Issue #247, opened 2023-06-27
Invalid debug offset invariant


guybedford:
Replication instructions in https://github.com/rustwasm/wasm-bindgen/pull/3483#issuecomment-1594273805.

The last instruction address from https://github.com/rustwasm/walrus/blob/main/src/module/debug/dwarf.rs#L214 ends up being after the subsequent function end address in https://github.com/rustwasm/walrus/blob/main/src/module/debug/dwarf.rs#L210.

Basic logging provides:

```
Instruction offset: 4834
Instruction offset: 4852
Instruction offset: 4859
Instruction offset: 4869
Instruction offset: 4884
Instruction offset: 4888
Instruction offset: 4898
Instruction offset: 4908
Instruction offset: 4909
Address conversion: 10337 -> 4407 (- 365 = 4042)
Function end: 4042
```

Where the broken invariant is then that `4909 < 4042` does not hold.

I tested the original PR and it also throws the exact same assertion so this was not an issue with the rebasing.

We maintain separate offset tables for the functions and instructions, and in this case, the instruction offsets are not present so we are falling back to the new `None` path in the `convert_address` function (which was not a part of the original Gimli implementation this came from that instead errored).

Separately, it's worth noting there's a language translation issue in the implementation where `Instrument` is used instead of `Instruction`, which might help code understanding to know.

Stefan-Hanke:
I'd like to tackle this - but beware I have not done system-level programming for quite some time. My main motivation is gaining experience in the rust ecosystem in general.

What I did so far:
- Replicated the issue locally - I was glad that the code-lldb VSCode extension worked without any other configuration on Windows, and that I found about that patch table in config.toml (neat stuff)
- Skimmed over DWARF v5 - never been that far down.
- ... yeah and then trying to understand line programs. The spec is a difficult read for me. I found some resources but yeah, need to bite through that.

I'll now go on with generating some ELF debug-built object files (I failed to cross-compile a _hello world_...) and then using [dwex](https://github.com/sevaa/dwex) in the hope of gaining insight.

I'd appreciate resources about line programs.

guybedford:
@Stefan-Hanke good luck! As I say I suspect the issue is somewhere in the offset handling code landed in the original PR that added this work, the `None` path in convert_address did seem a bit odd to me. If you have any questions at all in looking into this, happy to help in any way I can.

d3lm:
Is there any progress on this?

squillace:
[...]

PR #253, opened 2023-11-09
bugfix: DWARF transformation failure in some code

nokotan:
Resolves: #247 

In the original PR (#231), once the debug program sequence begins, it tries to convert any address in original wasm binaries, even if the address is not mapped in transformed wasm binaries.

That is based on the assumption that any consequence address is existent in transformed binary, if sequence-beginning address can be mapped to transformed one. However, the assumption seems not to be correct especially in inline functions.

In this PR, all addresses that convert function cannot converted are just ignored in DWARF line program converting process.

### DIFF
--- a/src/module/config.rs
+++ b/src/module/config.rs
@@ -87,6 +87,8 @@ impl ModuleConfig {
     /// itself!
     pub fn generate_dwarf(&mut self, generate: bool) -> &mut ModuleConfig {
         self.generate_dwarf = generate;
+        // generate_dwarf implies preserve_code_transform
+        self.preserve_code_transform = generate || self.preserve_code_transform;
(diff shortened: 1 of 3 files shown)


guybedford:
I just tested this against the previous bug in https://github.com/rustwasm/wasm-bindgen/pull/3483 and can confirm this resolves the issue.

d3lm:
Thanks a ton for this PR @nokotan 🙏

squillace:
Agreed! Thanks a ton, @nokotan !!! This is going to have a huge impact if we can use this now: https://marketplace.visualstudio.com/items?itemName=ms-vscode.wasm-dwarf-debugging

PR #257, opened 2023-12-29
Allow merging modules and skipping DWARF reading on load

gkgoat1:
I added this because, as far as I know, no other Rust WASM library can easily do this. I also need this for a project.

### DIFF
--- a/src/module/config.rs
+++ b/src/module/config.rs
@@ -208,6 +208,12 @@ impl ModuleConfig {
         Module::parse(wasm, self)
     }
 
+    /// Parses an in-memory WebAssembly file into an existing `Module` using this
+    /// configuration.
+    pub fn parse_into(&self, wasm: &[u8], pre: &mut Module) -> Result<()> {
+        Module::parse_in(wasm, self, pre)
+    }
+
     /// Parses a WebAssembly file into a `Module` using this configuration.
     pub fn parse_file<P>(&self, path: P) -> Result<Module>
     where
--- a/src/module/config.rs
+++ b/src/module/config.rs
@@ -18,6 +18,7 @@ pub struct ModuleConfig {
     pub(crate) on_parse:
         Option<Box<dyn Fn(&mut Module, &IndicesToIds) -> Result<()> + Sync + Send + 'static>>,
     pub(crate) on_instr_loc: Option<Box<dyn Fn(&usize) -> InstrLocId + Sync + Send + 'static>>,
+    pub(crate) no_read_dwarf: bool,
 }
 
 impl Clone for ModuleConfig {
@@ -32,6 +33,7 @@ impl Clone for ModuleConfig {
             skip_producers_section: self.skip_producers_section,
             skip_name_section: self.skip_name_section,
             preserve_code_transform: self.preserve_code_transform,
+            no_read_dwarf: self.no_read_dwarf,
 
             // ... and this is left empty.
             on_parse: None,
@@ -54,6 +56,7 @@ impl fmt::Debug for ModuleConfig {
             ref preserve_code_transform,
             ref on_parse,
             ref on_instr_loc,
+            ref no_read_dwarf,
         } = self;
 
         f.debug_struct("ModuleConfig")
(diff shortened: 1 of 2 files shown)


src/module/mod.rs
```diff
@@ -128,9 +129,21 @@ impl Module {
...
+        return Self::parse_in(wasm, &ModuleConfig::default(), self);
+    }
+
     fn parse(wasm: &[u8], config: &ModuleConfig) -> Result<Module> {
         let mut ret = Module::default();
         ret.config = config.clone();
+        Self::parse_in(wasm, config, &mut ret)?;
+        return Ok(ret);
+    }
+    fn parse_in(wasm: &[u8], config: &ModuleConfig, ret: &mut Module) -> Result<()> {
```
guybedford, line 143: Instead of having `ret` as the last arg, perhaps just make it the first `self` arg?

guybedford: Also, instead of taking a new `config`, we should probably use the same configuration, `ret.config` I think.

src/module/config.rs
```diff
@@ -69,6 +72,7 @@ impl fmt::Debug for ModuleConfig {
             .field("preserve_code_transform", preserve_code_transform)
             .field("on_parse", &on_parse.as_ref().map(|_| ".."))
             .field("on_instr_loc", &on_instr_loc.as_ref().map(|_| ".."))
+            .field("no_read_dwarf", no_read_dwarf)
```
guybedford, line 75: Let's stick to the existing pattern and call this `skip_debug_sections`.

src/module/functions/mod.rs
```diff
@@ -350,7 +350,7 @@ impl Module {
         // necessary and it's a bottleneck!
         let mut bodies = Vec::with_capacity(functions.len());
         for (i, (body, mut validator)) in functions.into_iter().enumerate() {
-            let index = (num_imports + i) as u32;
+            let index = (indices.num_fun_imports + i) as u32;
```
guybedford, line 353: I'd be curious to know the reason this is necessary - which is incorrect above - `funcs.arena.len()` or `functions.len()`? I'm also not particularly confident this covers all index changes necessary to support the feature, if you do want to land this it would help to provide a little more analysis / feedback on why you think this is enough to make this PR complete.

guybedford (changes requested):
Thanks for sharing. I'm open to landing the feature, but am not convinced we have covered all the necessary index changes.

I think to land we would absolutely need tests which have an existing module consisting of an import and a local of all core types, to which we then add new imports and locals of all core types. That would be the minimum I think to convince me this works comprehensively, especially given I don't have a completely thorough understanding of the codebase myself.

Completely understood if it's too much work.

Also, if you wanted to split out the `skip_debug_section` flag as its own PR that might also be easier to land independently.

src/module/functions/mod.rs
gkgoat1: I did this because `functions.len()` is the length of the current module's functions, which includes the total amount of imports being merged *and* functions previously in the module. Previously, `functions.len()` worked because `functions` started out empty, but with merging, `functions` can contain values before imports.

guybedford: Thanks for clarifying. Are there any other cases where imports being added later might affect indexing, or do you feel pretty confident this is the only one?

guybedford (commented):
Thanks for making the changes. This PR still adds a lot of public functions:

* ModuleConfig::read_dwarf
* ModuleConfig::skip_dwarf
* ModuleConfig::parse_into
* Module::merge_buffer
* Module::parse_in

Firstly, why does this configuration option need to have a function variant for both turning it on and off? We don't seem to do this for any other configuration options.

Then secondly, for the main method - can we not just design this to be a single new function? It would be much more preferable than introducing three new functions.

Finally this can only land with tests, so we will of course need to add that too.

gkgoat1:
By the way, did WABT change its syntax again?

guybedford:
CI should be passing here again now.

PR #266, opened 2024-06-25
feat!: upgrade wasmparser & wasm-encoder to v0.212.0

lwshang:
# Description

It is inevitable to introduce breaking changes so that we can make this project catch up with recent Wasm standard (and proposals).

## New types

* `RefType` is now a separate type which covers `Funcref` and `Externref` used to be in `ValType`
* `ElementItems` which covers both `Functions` and `Expressions`

## Types that have significant change

* `InitExpr` => `ConstExpr`
* `Element`: add `item: ElementItems`
* `Global`: add `shared: bool`
* `Memory`:
    * add `memory64: bool`
    * `u64` for `initial` and `maximum`
    * add `page_size_log2: Option<u32>`
* `Table`: 
    * add `table64: bool`
    * `u64` for `initial` and `maximum`

## Wasm Feature

In `module/mod.rs`:

Fix `wasm_features` from `wasmparser::WasmFeatures::default()` regardless of `config.only_stable_features`.

## Spec tests

* Update the spec-tests (WebAssembly/testsuite) submodule
* Reflect the latest set of Wasm proposals
* Now run all core tests and the multi-memory proposal tests
* Remove outdated special cases
* Collect and show all failed test cases instead of only the first one

## CI

Use the latest pre-compiled `wabt` and `binaryen`.

### DIFF
--- a/src/module/functions/local_function/mod.rs
+++ b/src/module/functions/local_function/mod.rs
@@ -44,7 +44,6 @@ impl LocalFunction {
     ///
     /// Validates the given function body and constructs the `Instr` IR at the
     /// same time.
-    #[allow(clippy::too_many_arguments)]
     pub(crate) fn parse(
         module: &Module,
         indices: &IndicesToIds,
@@ -70,7 +69,7 @@ impl LocalFunction {
             }),
         };
 
-        let result: Vec<_> = module.types.get(ty).results().to_vec();
+        let result: Vec<_> = module.types.get(ty).results().iter().cloned().collect();
         let result = result.into_boxed_slice();
 
         let controls = &mut context::ControlStack::new();
@@ -207,7 +206,6 @@ impl LocalFunction {
     }
 
     /// Emit this function's compact locals declarations.
-    #[allow(clippy::type_complexity)]
     pub(crate) fn emit_locals(
         &self,
         module: &Module,
@@ -301,7 +299,11 @@ fn block_param_tys(ctx: &ValidationContext, ty: wasmparser::BlockType) -> Result
     }
 }
 
-fn append_instruction(ctx: &mut ValidationContext, inst: Operator, loc: InstrLocId) {
+fn append_instruction<'context>(
+    ctx: &'context mut ValidationContext,
+    inst: Operator,
+    loc: InstrLocId,
+) {
     // NB. there's a lot of `unwrap()` here in this function, and that's because
     // the `Operator` was validated above to already be valid, so everything
     // should succeed.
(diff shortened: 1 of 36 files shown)


lwshang:
@guybedford 

All CI checks have passed! Could you please review and merge my PR when you get a chance?

Next up, I'll be working on the "memory64" feature in a new PR soon.

src/module/functions/local_function/mod.rs
```diff
@@ -1305,26 +1312,18 @@ fn append_instruction<'context>(
...
-        | Operator::Try { ty: _ }
-        | Operator::Catch { index: _ }
-        | Operator::Throw { index: _ }
-        | Operator::Rethrow { relative_depth: _ }
-        | Operator::Delegate { relative_depth: _ }
-        | Operator::CatchAll => {
-            unimplemented!("not supported")
-        }
+        // TODO: The remaining operators are for some on-going proposals
+        other => unimplemented!("unsupported operator: {:?}", other),
```
guybedford, line 1327: Rather than a generic `other` could we fully expand this but have all options throwing unsupported operator error? That way future upgrades won't miss additions to this list that may be important to know.

src/module/mod.rs
```diff
@@ -132,24 +133,23 @@ impl Module {
...
-            multi_value: true,
-            bulk_memory: !config.only_stable_features,
-            simd: !config.only_stable_features,
-            threads: !config.only_stable_features,
-            multi_memory: !config.only_stable_features,
-            ..WasmFeatures::default()
-        });
+
+        // TODO: how should we handle config.only_stable_features?
+        let wasm_features = WasmFeatures::default();
```
guybedford, line 138: For `only_stable`, it seems fine to just treat that as the default. That is, if `only_stable_features` is false we can enable all experimental features.

guybedford (commented):
This is great work, thank you for getting this one over the line. Just a couple of comments on the TODOs then it would be great to land and release this.

src/module/mod.rs
```diff
@@ -132,24 +133,24 @@ impl Module {
...
-            multi_value: true,
-            bulk_memory: !config.only_stable_features,
-            simd: !config.only_stable_features,
-            threads: !config.only_stable_features,
-            multi_memory: !config.only_stable_features,
-            ..WasmFeatures::default()
-        });
+
+        // For now we have the same set of wasm features
+        // regardless of config.only_stable_features.
```
guybedford, line 138: Perhaps we can add a note that new unstable features may be added in future under `only_stable_features: false`:

```suggestion
        // For now we have the same set of wasm features
        // regardless of config.only_stable_features. New unstable features
        // may be enabled under `only_stable_features: false` in future.
```

guybedford (approved):

PR #267, opened 2024-07-02
chore: add a clippy job in CI

lwshang:
## Description

Applied clippy suggestions.

Also bumped the dev-dependencies to the latest major versions.

### `gimli`

Now the only not yet updated dependency in Cargo.toml is `gimli`.
When trying to upgrade it, I got an deprecation warning about `Dwarf::borrow` and it requires more work to get rid of this warning message.
Also `gimli::read::Dwarf` is in the public API of `walrus`, upgrading major version of `gimli` is a breaking change.
```rs
pub struct ModuleDebugData {
    pub dwarf: read::Dwarf<Vec<u8>>,
}
```
Since we just released version 0.21.0, I believe it's not the best time to introduce a small change that would necessitate another major release soon.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,32 +9,21 @@ jobs:
       matrix:
         rust: [stable, beta, nightly]
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v4
       with:
         submodules: true
     - name: Install Rust
       run: rustup update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
-
-    # Build an up-to-date version of wabt since the releases don't always have
-    # all the features we want.
-    - uses: actions/checkout@v2
-      with:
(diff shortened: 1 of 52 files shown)


guybedford:
While this is not a BytecodeAlliance project specifically, BA projects as a general rule don't take clippy PRs or CI jobs.

That said I would be happy to apply improvements, but without anything that specifically calls out clippy or clippy conventions (ie the CI job and the clippy attribute).

guybedford:
> Since we just released version 0.21.0, I believe it's not the best time to introduce a small change that would necessitate another major release soon.

If you would like to look at the Gimli upgrade here, posting an immediate 0.22.0 would be fine by me actually, especially given that 0.21.0 has been out for not very long.

lwshang:
I see. I will make another PR which only upgrade dependencies and fix deprecation warnings.

PR #270, opened 2024-07-09
feat!: complete support memory64 proposal

lwshang:
# Types Changes

- Change the definition of `DataKind`
  - remove `ActiveData` and `ActiveDataLocation`

# Spec-tests changes

In the spec-tests, running `wast2json` to generate test cases was the first step. But it usually failed for proposals' wast files.
The reason is `wast2json` as a part of `wabt` project is slow to support new proposals.

Fortunately, the `wasm-tools` project recently provides a `json-from-wast` subcommand which support new proposals much better.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,9 @@ jobs:
         submodules: true
     - name: Install Rust
       run: rustup update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
+    - uses: cargo-bins/cargo-binstall@main
+    - name: Install wasm-tools (need json-from-wast subcommand)
+      run: cargo binstall wasm-tools
     - name: Install wabt
       run: |
         set -e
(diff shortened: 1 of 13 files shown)


.github/workflows/main.yml
```diff
@@ -14,6 +14,9 @@ jobs:
         submodules: true
     - name: Install Rust
       run: rustup update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
+    - uses: cargo-bins/cargo-binstall@main
+    - name: Install wasm-tools (need json-from-wast subcommand)
+      run: cargo binstall wasm-tools -y
```
guybedford, line 19: This can be quite a slow build to run on CI. It can be faster to just use the release directly.

See eg https://github.com/bytecodealliance/wasmtime/blob/main/.github/workflows/main.yml#L801.

lwshang: I'm using `cargo binstall` here which uses the pre-compiled binary from GitHub releases

guybedford: Ahh I missed that, nice.

PR #272, opened 2024-09-01
Support the Tail Call Proposal

CryZe:
This adds support for the Tail Call proposal.


guybedford:
Thanks for posting, let me know when it is ready to review with tests and I'd be happy to approve and release further.

CryZe:
I added roundtrip tests for both instructions. I'm not sure if this is sufficient or if you are looking for more tests. The test infrastructure also seems a bit confusing, so I wasn't entirely sure on how to make it fully work and what tests are generally needed here.

CryZe:
Nice, thank you :)

PR #280, opened 2024-11-08
Specify "shared" attribute of TableType to enable upgrade of wasm encoder to 0.213

burakemir:
wasm-encoder has a struct TableType which got a new attribute "shared".

The context is a proposal to share data among threads.

We set the attribute to false which means that the table is not to be shared.
This is the defensive setting and seems closest to status quo / ignoring the proposal.

In other words, this change only makes walrus compile with more recent versions and is not an attempt to do anything smart about the shared-everything-threads proposal.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,8 +29,8 @@ leb128 = "0.2.4"
 log = "0.4.8"
 rayon = { version = "1.1.0", optional = true }
 walrus-macro = { path = './crates/macro', version = '=0.22.0' }
-wasm-encoder = "0.212.0"
-wasmparser = "0.212.0"
+wasm-encoder = "0.213.0"
+wasmparser = "0.213.0"
 gimli = "0.26.0"
 
 [features]
--- a/src/module/imports.rs
+++ b/src/module/imports.rs
@@ -301,6 +301,7 @@ impl Emit for ModuleImports {
                             table64: table.table64,
                             minimum: table.initial,
                             maximum: table.maximum,
+                            shared: false,
                         })
                     }
                     ImportKind::Memory(id) => {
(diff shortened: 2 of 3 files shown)


guybedford (changes requested):
In the wasmparser side we should add a read of this property and throw if it is found if we are not handling this state.

Alternatively we should include it in the representation.

burakemir:
Please help me understand better what you mean.

wasmparser already has validation, in the following sense: shared=true is only ever permitted if the "shared everything threads" proposal is enabled: https://github.com/burakemir/wasm-tools/blob/3e0dda0a4736848e48b699d064e2173375eb6560/crates/wasmparser/src/validator/core.rs#L771

So I am pretty confident that shared=false is a good, safe value, not only to get walrus to compile, but also if walrus-generated WASM is used in places where shared=true may be encountered. 

The [shared everything proposal](https://github.com/WebAssembly/shared-everything-threads/blob/main/proposals/shared-everything-threads/Overview.md) does change a lot of things, and supporting WASM transformations in a shared everything world probably benefits from a careful look at, well, everything, and potentially some redesign. 

No matter what happens with respect to supporting shared-everything, a small change like this PR seems necessary just to catch up. I have another PR that adds all the new atomic instructions (as unsupported operations, also just to get things to build again).

guybedford:
I'm asking one of two things:

1. Fully support the `shared` state in Walrus in both parsing and serializing
2. Throw if the `shared = true` state is found during parsing

You are welcome to do either, but we absolutely must do one to avoid incorrectly interpreting Wasm binaries.

guybedford:
I suppose shared everything is disabled in the features by default so this is fine for now, we just need to watch if that default changes.

PR #281, opened 2024-11-12
Enable relaxed SIMD by default

daxpedda:
This enabled the relaxed SIMD proposal, which is in phase 5, by default. `wasm-bindgen` currently fails without this addition when compiling Wasm modules containing relaxed SIMD instructions, which are stable since Rust v1.82.

A new point release would be appreciated as well!

### DIFF
--- a/src/ir/mod.rs
+++ b/src/ir/mod.rs
@@ -605,6 +605,13 @@ pub enum Instr {
         /// The table which `func` below is indexing into
         table: TableId,
     },
+
+    /// Various relaxed SIMD instructions.
+    RelaxedSimd {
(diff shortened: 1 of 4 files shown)


daxpedda:
> EDIT: I pushed an incomplete version here: [main...dragostis:walrus:relaxed-simd](https://github.com/rustwasm/walrus/compare/main...dragostis:walrus:relaxed-simd)

Ah, just saw it after I finished :sweat_smile:, thanks nevertheless!

I went ahead and renamed `RelaxedSimd` to `TernOp`, which seems more appropriate considering no special handling is required.

PR #282, opened 2024-11-13
Add missing ops to enable upgrade to 0.214

burakemir:
Adds cases for the added shared-everything-threads instructions.

These instructions are part of the shared everything threads [proposal](https://github.com/bytecodealliance/wasm-tools/pull/1664) and were added to wasmparser in https://github.com/bytecodealliance/wasm-tools/commit/bf282bb1054a58b7989b112de522674999d4af50 

This change treats all of them as unsupported, with the intent to enable upgrading wasmparser and wasm-encoders tools dependencies in lock step.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,8 +29,8 @@ leb128 = "0.2.4"
 log = "0.4.8"
 rayon = { version = "1.1.0", optional = true }
 walrus-macro = { path = './crates/macro', version = '=0.22.0' }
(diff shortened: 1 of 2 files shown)


burakemir:
To clariy, this change is necessary if we want walrus to build with wasmparser >=0.214

An alternative approach is to give up on enumerating all instructions and accept a catch-all case in src/module/functions/local_function/mod.rs

PR #284, opened 2024-11-29
Don't emit empty DWARF sections

daxpedda:
This change prevents empty custom sections for debug information to be generated when no debug information is available. `fallible-iterator` is not a new dependency, it is a dependency of `gimli`, but is not re-exported.

Not only are a bunch of empty custom sections undesirable, seemingly `wasm-opt` had issues parsing those. I'm unsure if this is a bug or not, I don't know if completely empty DWARF sections are valid or not.

This will enable `wasm-bindgen` to stop throwing away the debug information if its available.

See https://github.com/rustwasm/wasm-bindgen/pull/3876.

### DIFF
--- a/src/module/debug/mod.rs
+++ b/src/module/debug/mod.rs
@@ -115,10 +115,12 @@ impl Emit for ModuleDebugData {
         sections
             .for_each(
                 |id: SectionId, data: &write::EndianVec<LittleEndian>| -> Result<()> {
-                    cx.wasm_module.section(&wasm_encoder::CustomSection {
-                        name: id.name().into(),
-                        data: data.slice().into(),
-                    });
+                    if !data.slice().is_empty() {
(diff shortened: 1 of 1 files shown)


philipc:
While empty DWARF sections shouldn't be a problem, it's pointless to emit them. This applies even if some DWARF sections have data and others don't. So I think it would be better to skip the following if `data` is empty: https://github.com/rustwasm/walrus/blob/27f6b6266e4625443a95dd54f6c772498215536a/src/module/debug/mod.rs#L118-L121

daxpedda:
> While empty DWARF sections shouldn't be a problem, it's pointless to emit them. This applies even if some DWARF sections have data and others don't. So I think it would be better to skip the following if `data` is empty:

Ha, this is actually what I had implemented initially.
Will do that then, thank you!

